Issue #25, opened 2020-02-24
Recommend using `npx` over installing react-native-cli globally


tido64:
According to React Native CLI [documentation](https://github.com/react-native-community/cli/blob/master/docs/init.md), you're not supposed to install `react-native-cli` globally any longer. The recommended approach is to use `npx`, e.g.:

```bash
npx react-native init MyApp
# or
npx react-native@^0.60 init MyApp
```

chrisglein:
@kikisaints Can you try the getting started steps using npx and confirm whether we're ready to change the recommended steps or not yet? And if we are, updated the documentation.

kikisaints:
This has been addressed as of #33 and #31 .

PR #30, opened 2020-03-05
Update getting-started.md to use npx and removed global react native install

kikisaints:
This is a minor update to the getting started guide that prefixes the react-native init with npx. This change also removes the global install per the new 0.60 instructions.

This addresses #25 .

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -70,18 +70,12 @@ For information around how to set up React Native, see the [React Native Getting
   <div id="stable">
 The latest and recommended release to use.
 
-## Install React native
-
-```
-npm install -g react-native-cli
-```
-
 ## Install React Native for Windows
 
 Remember to call this ``react-native init`` in the directory you want your project to live.
 
 ```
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -70,18 +70,12 @@ For information around how to set up React Native, see the [React Native Getting
   <div id="stable">
 The latest and recommended release to use.
 
-## Install React native
```
harinikmsft: We should have a step to install react native, just point to React Native getting stated guide from that section.

kikisaints: From my understanding of npx, it also installs React Native for you, just locally instead of globally. @acoates-ms is that correct?

acoates-ms: Yeah, npx will install react-native to a local cache similar to install -g, but unlike install -g, it wont pollute your global state.

harinikmsft: Very cool. The update looks good in that case.

PR #40, opened 2020-03-19
Updating the SampleNativeModule to generate winmds

jonthysell:


harinikmsft:
cc: @kikisaints - may need to manually publish so the site is updated?

PR #87, opened 2020-05-04
Remove out of date "implemented" column

### LINKED ISSUE #86: APi Parity "implemented" column for modules is inaccurate
https://microsoft.github.io/react-native-windows/docs/parity-status#supported-community-modules

chrisglein:
The dates were wrong on the column of implemented modules. Also, it's not terrible helpful, so I removed it.
These modules were also all described as being migrated out, which isn't correct for all of them. Broke into two tables, one for lean core migrations, the other for community modules.
Long term those should be tracked with a [directory](https://github.com/microsoft/react-native-windows/issues/3718), but for now while the list is short this is still helpful.

Fixes #86 86

### DIFF
--- a/docs/parity-status.md
+++ b/docs/parity-status.md
@@ -14,19 +14,25 @@ A few methods or props may be missing on some types and we are actively working 
 If you encounter an unsupported API that should be tracked, please [submit an issue](https://github.com/microsoft/react-native-windows/issues/new/choose) to let us know.
 
 ## Supported Community Modules
-We are in the process of ejecting the components and modules that are not part of React Native Lean Core into their respective [community repos](https://github.com/react-native-community).
+We are in the process of ejecting the components and modules that are not part of React Native [Lean Core](https://github.com/facebook/react-native/issues/23313) into their respective [community repos](https://github.com/react-native-community).
 
-The following were completed recently:
+The following have been migrated out:
 
-| Name | Status | Implemented |
-|:-|:-|-:|
-| <ins>[react-native-webview](https://www.github.com/react-native-community/react-native-webview)</ins> | **Completed** |Mar 17, 2020 |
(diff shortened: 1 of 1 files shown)


chrisglein:
@kikisaints @harinikmsft @stmoy Thoughts on how we want to represent this information? I both removed the incorrect information here but also explored how we may want to represent the difference between lean core moves and community modules.

harinikmsft:
> @kikisaints @harinikmsft @stmoy Thoughts on how we want to represent this information? I both removed the incorrect information here but also explored how we may want to represent the difference between lean core moves and community modules.

@chrisglein - I like what you have done here. It will be a challenge to keep this page updated as we add more modules, but that's a good problem to have in the near term. In the long term, once we have some form of critical mass in supported modules, we could simply point to the [reactnative.directory](reactnative.directory) site with a filter by Windows.

In the short term, we need a similar page for macOS as well. Let me work on adding that.

harinikmsft:
@chrisglein FYI - I just merged some changes that fixed the broken link stuff that your CI is hitting.

chrisglein:
> @chrisglein I like what you have done here. It will be a challenge to keep this page updated as we add more modules, but that's a good problem to have in the near term. In the long term, once we have some form of critical mass in supported modules, we could simply point to the [reactnative.directory](reactnative.directory) site with a filter by Windows.
 
Agreed this doesn't scale up well as we grow. But for where we are right now I think it answers some common questions while we wait to get more representation in the directory. So I think we do this for awhile... and then delete it when we can point people somewhere else.

> In the short term, we need a similar page for macOS as well. Let me work on adding that.

Yeah, we'll want this same sort of thing with macOS, agreed. This page is under the Windows section, so we'll need an additional page under the macOS docs.

PR #97, opened 2020-05-12
Big visual update for Build

kikisaints:
This is a large visual update to the site. This also includes adding the call to action buttons on the main page to the getting started guides for both windows and mac.


harinikmsft:
I am merging this now to ensure the site gets updated. @jonthysell - feel free to provide feedback, Kiki can address those in subsequent PRs.

PR #99, opened 2020-05-13
Update getting-started.md

csbarnes:
If `npx react-native run-windows` is not run from an elevated Command Prompt it will fail.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -47,7 +47,7 @@ npx react-native-windows-init --overwrite
 
 - Without Using Visual Studio
 
-  In your React Native Windows project directory, run:
+  In your React Native Windows project directory, from an elevated Command Prompt run:
 
   ```
   npx react-native run-windows


chrisglein:
@asklar has done some review of where we need elevation. Adding him to requested reviewers to comment.

asklar:
> @asklar has done some review of where we need elevation. Adding him to requested reviewers to comment.

Yes this should only be the case for _some_ cases. The short of it:
- The first time you run it, we may have to enable developer mode which requires elevation. We prompt for elevation now if we need it so it doesn't fail.
- For debug builds on C++ projects, we deploy from layout so registering the APPX is not necessary which lets us avoid elevation
- For release builds, we'll prompt for elevation if we need to in order to install the app signing certificate and uninstall the existing app.

I also have a PR out for a script to check and install any dependencies that may be missing which will alleviate some of this too. See https://github.com/microsoft/react-native-windows/pull/4876

PR #114, opened 2020-05-18
 fix text in Resources page

luism3861:
this PR fix text in Resources Page.


harinikmsft:
@luism3861 - Thanks for the submission!

PR #120, opened 2020-05-20
remove manual Pods installation from macOS guide

Simek:
This small PR removes the manual Pods installation step from the latest Getting Started guide for the macOS. 

Refs: microsoft/react-native-macos#366


harinikmsft:
cc: @tom-un . Can you take a look at this PR?

PR #126, opened 2020-05-21
Minor proposals on lines 204 & 508

kant:


kikisaints:
Thank you for all the typo fixes @kant !

PR #135, opened 2020-06-05
Added resources to the resources page

kikisaints:
This adds fun, stylized cards on the resources page that link out to repos related to the RNWM effort.


website/pages/en/resources.js
```diff
@@ -89,36 +74,77 @@ class Resources extends React.Component {
     const VideoCardList = () => (
       <Section background="tint">
         <div className="content">
-          <MarkdownBlock>{textContent.introtext}</MarkdownBlock>
+          <h1 style={{marginTop: -35}}>Resources</h1>
+          <p>The follow is a list of resources related to React Native for Windows and macOS to help you get started or to better accelerate your development in this area.</p>
```
harinikmsft, line 78: Feels like this page is more highlighting the various offerings from Microsoft in the React Native space rather than specifically being about resources (like docs, videos, guides etc.,) and they are not specific to RN for Windows/Mac only either. So, maybe change header/title/description to reflect that a bit more? Something like - "The following is a suite of offerings from Microsoft in the React Native space to help you target Desktop platforms as well as accelerate your development efficiency with React Native".

website/pages/en/resources.js
```diff
@@ -89,36 +74,77 @@ class Resources extends React.Component {
...
+                    imgurl="./img/homepage/vs_code_logo.png"
+                    cardlink="https://github.com/Microsoft/vscode-react-native"
+                    videodisc="React Native extension for VS Code enables you to debug your code and quickly and run react-native commands from the command palette."/>
+                    <ResourceCardItem
+                      videolength="Repo"
+                      videotitle="App Center SDK for React Native"
+                      videotype="Github Repo"
+                      imgurl="./img/homepage/appcenter_logo.png"
+                      cardlink="https://github.com/microsoft/appcenter-sdk-react-native"
+                      videodisc="React Native extension for VS Code enables you to debug your code and quickly and run react-native commands from the command palette."/>
```
harinikmsft, line 138: The videodesc field of this one is a copy of VSCode item from previous line. Should update with Appcenter specific description.

website/pages/en/resources.js
kikisaints: Hmm, sidebars are only enabled for markdown-only files. I can remove the styling/cards here and create a separate markdown file - then the resources will be versioned too, would that work? Only drawback would be that everything would then go into a markdown table which is a bit less readable...

website/pages/en/resources.js
kikisaints: thanks! fixed.

website/pages/en/resources.js
harinikmsft: I really like the styling cards, is there a way to just embed these as html blocks inside a markdown file so we can preserve the format? We do similar things in the landing pages of the repos to embed the top header items as html content.

kikisaints:
merging for now, will create an issue for this page's mobile responsiveness later.

PR #144, opened 2020-06-11
WinUI3 docs

asklar:
This adds a doc for WinUI 3 support, I'll continue to update it as we make more progress on features and thinking about issues like native modules support.


docs/winui3.md
```diff
@@ -0,0 +1,34 @@
+---
+id: winui3
+title: WinUI 3
+---
+
+# WinUI 3 Support in RNW
+
+We are working on adding **WinUI 3** support for RNW and have made some good progress towards an alpha.
```
harinikmsft, line 8: Can you add some preamble on what WinUI 3 is (perhaps with links to the WinUI3, project reunion repos) and why we are doing this so there is context for people coming new?

docs/winui3.md
```diff
@@ -0,0 +1,34 @@
...
+## Current status
+
+As of June 2020, we support the **WinUI 3.0 Alpha** release for:
+
+* Creating a new RNW app C++/WinRT app. 
+* The Playground app is able to build a `WinUI 3` Configuration (you can change this in VS). This corresponds to a `Debug` build but it targets WinUI 3.
+
+## Not yet supported
+
+* **`Release` builds** - this is an in-development feature so it really isn't intended for release/publishing yet. You'll see build breaks when switching to Release since the `WinUI3` is based on the `Debug` configuration.
```
harinikmsft, line 19: Do we have issues in the main repo tracking these items? It might be useful to point to them to continue the conversation/track progress

docs/winui3.md
harinikmsft: You could borrow from [this doc](https://microsoft.github.io/react-native-windows/docs/win10-compat) or maybe even add the content from that doc into yours if it is meaningful.

docs/winui3.md
asklar: Not yet - we'll probably file these as we do planning for 63 in the coming weeks and we can update this doc then? :)

PR #145, opened 2020-06-11
[0.63] Autolinking docs

jonthysell:
This change adds documentation to cover the autolinking feature added in RNW 0.63. See https://github.com/microsoft/react-native-windows/pull/5044

We cannot check this in until the website version bumps to 62.


docs/config.md
```diff
@@ -0,0 +1,175 @@
...
+
+|  Tag   | Description |
+|:-------|:------------|
+| *auto* | Item is always calculated by config. An override file should NEVER provide it. |
+| *req*  | Item is required. If an override file exists, it MUST provide it. If no override file exists, config will try to calculate it. |
+| *opt*  | Item is optional. If an override file exists, it MAY provide it. If no override file exists, config may try to calculate it. |
+
+## projectConfig
+
+`react-native config` will generate the following JSON for app projects that have a windows implementation, as a target for auto-linking. This is done heurestically, so if the result isn't quite correct, app developers can provide a manual override file: `react-native.config.js`.
```
asklar, line 22:
```suggestion
`react-native config` will generate the following JSON for app projects that have a Windows implementation, as a target for auto-linking. This is done heuristically, so if the result isn't quite correct, app developers can provide a manual override file: `react-native.config.js`.
```

docs/config.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Project Object Fields:
+
+The top-level `project` has the following fields:
+
+| Field | Type | Tag  | Description |
+|:------|:----:|:----:|:------------|
+| projectFile | string | req | Relative path to the VS project file under *sourceDir*, ex: *'MyApp\MyApp.vcxproj'* for *'c:\path\to\my-app\windows\MyApp\MyApp.vcxproj'* |
+| projectName | string | auto | Name of the project, determined from *projectFile*, ex: *'MyApp'* |
+| projectLang | string | auto | Language of the project, cpp or cs, determined from *projectFile* |
```
asklar, line 59:
```suggestion
| projectLang | string | auto | Language of the project, `cpp` (for C++) or `cs` (for C#), determined from *projectFile* |
```

docs/config.md
```diff
@@ -0,0 +1,175 @@
...
+        projectFile: 'MyApp\\MyApp.vcxproj',
+      },
+    },
+  },
+};
+```
+
+## dependencyConfig
+
+`react-native config` will generate the following JSON for each native module dependency under node_modules that has a windows implementation, in order to support auto-linking. This is done heurestically, so if the result isn't quite correct, native module developers can provide a manual override file: `react-native.config.js`.
```
asklar, line 80:
```suggestion
`react-native config` will generate the following JSON for each native module dependency under `node_modules` that has a Windows implementation, in order to support auto-linking. This is done heuristically, so if the result isn't quite correct, native module developers can provide a manual override file: `react-native.config.js`.
```

docs/native-modules-autolinking.md
```diff
@@ -0,0 +1,64 @@
...
+> Each platform defines its own [`platforms`](https://github.com/react-native-community/cli/blob/master/docs/platforms.md) configuration. It instructs the CLI on how to find information about native dependencies. This information is exposed through the [`config`](https://github.com/react-native-community/cli/blob/master/docs/commands.md#config) command in a JSON format. It's then used by the scripts run by the platform's build tools. Each script applies the logic to link native dependencies specific to its platform.
+
+The information provided by `config` is described in [React Native Config Schema](config.md).
+
+## Autolinking process
+
+Autolinking is performed automatically as a part of the `run-windows` command:
+
+1. At build time, autolinking is performed first, before `msbuild.exe` is invoked and the build actually started. It uses the information provided by `config` to both generate and modify certain native files consumed by your app project.
+    1. The `AutolinkedNativeModules.g.targets` file contains the necessary references to the dependency projects that must be built.
```
asklar, line 32: Where do these files get generated, relative to the solution source directory?
Ideally they go into the Generated Files folder

PR #160, opened 2020-07-09
Update documentation to include the new properties of the redbox handler

acoates-ms:
With the 0.63 merge we added some more properties to the Redbox error information available to native error handling.


kikisaints:
@acoates-ms CI checker is saying you have some issues - I'm perfectly fine with a squash merge, but can you check the links first?

acoates-ms:
@kikisaints I just went ahead and got the CI clean, since its erroring out on everyones PRs, figured we should fix that.

PR #163, opened 2020-07-11
Update native-modules.md

LionelHoudelier:
false first using line

### DIFF
--- a/docs/native-modules.md
+++ b/docs/native-modules.md
@@ -56,7 +56,7 @@ Here is a sample native module written in C# called `FancyMath`. It is a simple 
 _FancyMath.cs_
 
 ```csharp
-using System;
+using Microsoft.ReactNative;
 using Microsoft.ReactNative.Managed;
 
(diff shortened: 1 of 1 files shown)


kikisaints:
@stmoy can you take a look at this?

stmoy:
CC: @jonthysell 

This doc is a companion to the code, here: https://github.com/microsoft/react-native-windows-samples/blob/c4e072a20a67056f188e9a6ca502da5a7e60c8e6/samples/NativeModuleSample/csharp/windows/NativeModuleSample/FancyMath.cs#L4

This file seems to require System and not Microsoft.ReactNative. 

@LionelHoudelier - to complete the proposed change, we'd probably also want to update the NativeModuleSample C# sample code and ensure that it builds. Assuming it does, I'm good with this change but we shouldn't update one without updating (and verifying) both.

jonthysell:
NativeModuleSample currently targets 0.62. Is this using a necessary change to build in 0.63?

PR #185, opened 2020-08-12
Update native-modules-setup.md

stratospheres:
Fixed a broken link to the "project structure" docs. Had ".md" tacked on the end, which was the problem.

### DIFF
--- a/docs/native-modules-setup.md
+++ b/docs/native-modules-setup.md
@@ -127,7 +127,7 @@ We're going to add several React Native Windows projects to your solution. So to
 
 Now we're going to add all of the following React Native Windows projects to that `ReactNative` folder. All of these projects are located under the `node_modules\react-native-windows` directory in the root of your `react-native-my-library` project directory.
 
->*For more details about what these projects do, see [Project Structure](https://github.com/microsoft/react-native-windows/blob/master/docs/project-structure.md).*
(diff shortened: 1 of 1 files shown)


stratospheres:
Appears that the "check for broken links" step in CI is broken. Doesn't understand that the .md links don't really need the .md part in the URL. Given what I'm seeing here, this might indicate that other .md links throughout the docs are broken too, unless I'm missing something.

chrisglein:
> Appears that the "check for broken links" step in CI is broken. Doesn't understand that the .md links don't really need the .md part in the URL. Given what I'm seeing here, this might indicate that other .md links throughout the docs are broken too, unless I'm missing something.

@asklar Can you look into the broken link issue?

asklar:
The link with the .md is the correct one. The one without gives a 404

jonthysell:
The check is operating as expected - .md is required https://github.com/microsoft/react-native-windows/blob/master/docs/project-structure is a dead link.

stratospheres:
That's weird. It definitely was 404 *with the .md* before, and 200 without it, as my change suggests. It seems that @chrisglein saw the same thing (or at least I think so, given the PR approval). 

Did something change with the hosting? I'm 100% sure that this wasn't the behavior at that point...

stratospheres:
Maybe germane to the question, maybe not...

image

PR #208, opened 2020-08-25
Update native-modules-async.md to reflect `Task<T>` support in 0.63

dannyvv:
Update docs to use Task<T>

### DIFF
--- a/docs/native-modules-async.md
+++ b/docs/native-modules-async.md
@@ -25,122 +25,94 @@ NativeModules.SimpleHttpModule.GetHttpResponse('https://microsoft.github.io/reac
 
 ### SimpleHttpModule in C#
 
-Let's start with the asynchronous native method which performs the HTTP request:
+The native module support for C# supports the common async patterns established in C# using `async`, `await` and `Task<T>`.
 
-```csharp
-static async Task GetHttpResponseAsync(string uri)
-{
-  // Create an HttpClient object
-  var httpClient = new HttpClient();
+To expose the module to javascript you need to declare a C# class. To indicate it should be exposed to javascript, you annotate with an [attribute](https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/attributes) like:
 
-  // Send the GET request asynchronously
-  var httpResponseMessage = await httpClient.GetAsync(new Uri(uri));
-
-  var statusCode = httpResponseMessage.StatusCode;
-  var content = await httpResponseMessage.Content.ReadAsStringAsync();
-
-  // TODO: How to return the result?
-}
-```
-
-The `GetHttpResponseAsync` method is pretty straight-forward at this point, it takes a `string` uri and "returns" a `Task` (which is to say, the method is asynchronous and doesn't actually return a value when it's done).
-
-> If you're not familiar with writing asynchronous C# code, see [Call asynchronous APIs in C# or Visual Basic](https://docs.microsoft.com/en-us/windows/uwp/threading-async/call-asynchronous-apis-in-csharp-or-visual-basic).
-
-Inside `GetHttpResponseAsync`, we see it:
-1. Creates a `HttpClient`.
-2. Asynchronous calls the `GetAsync` method to make an HTTP request for the uri.
-3. Parses the status code out of the returned `HttpResponseMessage` object.
-4. Asynchronously parses the content out of the returned `HttpResponseMessage` object.
-
-Now we have `statusCode` and `content`, but what do we do with it? How do we call this method from JS, and how do we get the result back to the JS?
-
-Let's pause this and start building our native module:
-
-```csharp
+```c#
 namespace NativeModuleSample
 {
   [ReactModule]
   class SimpleHttpModule
   {
-    [ReactMethod]
-    public void GetHttpResponse(string uri, ReactPromise<JSValue> promise)
-    {
-    }
+    // Methods go here.
   }
 }
 ```
 
-Here we simply define `SimpleHttpModule` with an empty `GetHttpResponse` method.
+This makes an object available to JavaScript via the expression `NativeModules.SimpleHttpModule`. By default it will use the name of the class as the name in JavaScript. If you don't want the name of the class to match the name in JavaScript, i.e. you want to access the module the expression `NativeModules.CustomModule`. you can pass a custom name like:
 
-Notice the method itself is `void` and that the last parameter in the signature is of type `ReactPromise<JSValue>`. This indicates to React Native Windows that we want a promise-based method in JS, and that the expected return value of a success is of type `JSValue`.
-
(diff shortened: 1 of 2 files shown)


dannyvv:
@qmatteoq: FYI

qmatteoq:
I think we should have different docs based on the version, since Task are supported only starting from 0.63 (which is still in preview currently).

dannyvv:
> I think we should have different docs based on the version, since Task are supported only starting from 0.63 (which is still in preview currently).

My understanding is that docs in the root of the repo are vnext per [versions](https://microsoft.github.io/react-native-windows/versions) page, and there are versioned archives 
 [here](https://github.com/microsoft/react-native-windows-samples/tree/master/website/versioned_docs)

qmatteoq:
Ok fine by me. Once this PR gets approved, I'll make sure to include the original guidance in the archived docs. Thanks!

docs/native-modules-async.md
```diff
@@ -25,122 +25,94 @@ NativeModules.SimpleHttpModule.GetHttpResponse('https://microsoft.github.io/reac
...
-    public void GetHttpResponse(string uri, ReactPromise<JSValue> promise)
-    {
-    }
+    // Methods go here.
   }
 }
 ```
 
-Here we simply define `SimpleHttpModule` with an empty `GetHttpResponse` method.
+This makes an object available to JavaScript via the expression `NativeModules.SimpleHttpModule`. By default it will use the name of the class as the name in JavaScript. If you don't want the name of the class to match the name in JavaScript, i.e. you want to access the module the expression `NativeModules.CustomModule`. you can pass a custom name like:
```
asklar, line 43:
```suggestion
This makes an object available to JavaScript via the expression `NativeModules.SimpleHttpModule`. By default the JavaScript name will match the C# class name. If you don't want the name of the class to match the name in JavaScript, i.e. you want to access the module via the expression `NativeModules.CustomModule`. you can pass a custom name like:
```

docs/native-modules-async.md
```diff
@@ -25,122 +25,94 @@ NativeModules.SimpleHttpModule.GetHttpResponse('https://microsoft.github.io/reac
...
-  // Create an HttpClient object
-  var httpClient = new HttpClient();
+The native module support for C# supports the common async patterns established in C# using `async`, `await` and `Task<T>`.
 
-  // Send the GET request asynchronously
-  var httpResponseMessage = await httpClient.GetAsync(new Uri(uri));
-
-  var statusCode = httpResponseMessage.StatusCode;
-  var content = await httpResponseMessage.Content.ReadAsStringAsync();
+To expose the module to javascript you need to declare a C# class. To indicate it should be exposed to javascript, you annotate with an [attribute](https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/attributes) like:
```
asklar, line 30:
```suggestion
To expose the module to JavaScript you need to declare a C# class. To indicate it should be exposed to JavaScript, you annotate with an [attribute](https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/attributes) like:
```

docs/native-modules-async.md
```diff
@@ -25,122 +25,94 @@ NativeModules.SimpleHttpModule.GetHttpResponse('https://microsoft.github.io/reac
...
+```c#
+  [ReactModule("CustomModule")]
+  class SimpleHttpModule
+  { 
+    ...
+  }
+```
 
-The `promise` object is our interface for handling the promise and marshalling a result to the JS. To do so we simply call `promise.Resolve()` with the result object (if the operation was a success) or `promise.Reject()` with an error (if the operation failed).
+Now we want to expose the method that performs the http request. It is recommended, and it is the default, to write these functions asynchroneously. Writing async code in C# is pretty straight-forward and intuitivie with the `async` and `await` keywords and the `Task<T>` types.
```
asklar, line 53:
```suggestion
Now we want to expose the method that performs the HTTP request. It is recommended, and it is the default, to write these functions asynchronously. Writing async code in C# is pretty straight-forward and intuitive with the `async` and `await` keywords and the `Task<T>` types. 
```

docs/native-modules-async.md
```diff
@@ -25,122 +25,94 @@ NativeModules.SimpleHttpModule.GetHttpResponse('https://microsoft.github.io/reac
...
-  promise.Resolve(resultObject);
-}
+      var statusCode = httpResponseMessage.StatusCode;
+      var content = await httpResponseMessage.Content.ReadAsStringAsync();
+      
+      return content;
 ```
 
-At the bottom of the method, we simply build the result object to be returned to JS, and pass it to `promise.Resolve()`. That's it for `GetHttpResponseAsync` - if the method execution gets to the end without any problems, it will resolve the promise, which marshals the result back to the JS.
+Code code:
```
asklar, line 81: code code?

docs/native-modules-async.md
```diff
@@ -25,122 +25,94 @@ NativeModules.SimpleHttpModule.GetHttpResponse('https://microsoft.github.io/reac
...
+1. Creates a `HttpClient`.
+2. Asynchronous calls the `GetAsync` method to make an HTTP request for the uri.
```
asklar, line 83:
```suggestion
2. Asynchronously calls the `GetAsync` method to make an HTTP request for the URI.
```

docs/native-modules-async.md
```diff
@@ -25,122 +25,94 @@ NativeModules.SimpleHttpModule.GetHttpResponse('https://microsoft.github.io/reac
...
-
-> **Important:** This example shows the minimum case, where you don't handle any errors within `GetHttpResponseAsync`, but you're not limited to this. You're free to detect error conditions within your code and call `promise.Reject()` yourself with (more useful) error messages at any time. However you should *always* include this final handler, to catch any unexpected and unhandled exceptions that may occur, especially when calling Windows APIs. Just be sure that you only call `Reject()` once and that nothing executes afterwards.
+But wait, we only discussed the success path, what happens if `GetHttpResponse` doesn't succeed? We don't handle any exceptions in this example, so if an exception is thrown, how do we marshal an error back to the JS? That is actually take care of for you by the framework, any exception in the task will be marshalled to the JavaScript side as a javascript exception.
```
asklar, line 115:
```suggestion
But wait, we've only discussed the success path, what happens if `GetHttpResponse` doesn't succeed? We don't handle any exceptions in this example, so if an exception is thrown, how do we marshal an error back to JavaScript? That is actually taken care of for you by the framework: any exception in the task will be marshalled to the JavaScript side as a JavaScript exception.
```

dannyvv:
> Ok fine by me. Once this PR gets approved, I'll make sure to include the original guidance in the archived docs. Thanks!

When we cut 0.63 that should be an automatic operation, it will auto-create a 0.63 version, and make that the default on the website. Jon Thysel exlained all the details in a brownbag a while ago: https://msit.microsoftstream.com/video/d72ca4ff-0400-96d0-221b-f1ea8b25bbe3

docs/native-modules-async.md
dannyvv:
```suggestion
The code takes the following steps:
```

PR #215, opened 2020-08-27
Add docs for 0.63 csharp codegen.

dannyvv:


docs/native-modules-csharp-codegen.md
```diff
@@ -0,0 +1,94 @@
+---
+id: native-modules-nuget
+title: Compile time code generaton for C#
+---
+
+>**This documentation and the underlying platform code is a work in progress.**
+
+In previous versions of React-Native-Windows the C# interop operated by reflection. In 0.63 we improved this by adding a compile time code generation.
```
NickGerleman, line 8:
```suggestion
In previous versions of React Native for Windows, code generation for C# modules was performed using reflection. In 0.63 we improved this by adding a compile time code generation.
```

docs/native-modules-csharp-codegen.md
```diff
@@ -0,0 +1,94 @@
...
+
+There are several benefits to this approach:
+* The cost of analyzing and reflecting over your code doesn't have to happen at each application startup, it only happens once at build time.
+* Improved error reporting. Because the generator operates on the source code it can provide file and line information when an error is detected. 
+* You will discover all errors at compile time, rather than when the code is execersized.
```
asklar, line 13:
```suggestion
* You will discover all errors at compile time, rather than when the code is exercised.
```

docs/native-modules-csharp-codegen.md
```diff
@@ -0,0 +1,94 @@
...
+
+# How to use it
+## New projects
+If you create a new project from template (the recommended way), there is nothing you need to do to take advantage of this new feature, it is enabled by default.
+
+## Return to use reflection
+If prefer to use the previous reflection logic you can do so by edditting the msbuild project and add:
```
asklar, line 20:
```suggestion
If prefer to use the previous reflection logic you can do so by editing the MSBuild project (the `.vcxproj` or `.csproj` file) and add:
```

NickGerleman, line 20:
```suggestion
If you run into issues with compile-time code generation, you can return to previous behavior by editing the MSBuild project to add:
```

docs/native-modules-csharp-codegen.md
```diff
@@ -0,0 +1,94 @@
...
+
+```diff
+   <PropertyGroup>
+     ...
++    <ReactNativeCodeGenEnabled>true</ReactNativeCodeGenEnabled>
+   </PropertyGourp>
+```
+Next you will have to add the reflection based generation. Open `ReactPackageProvider.cs` and make the folowing changes:
```
asklar, line 28:
```suggestion
Next you will have to add the reflection-based generation. Open `ReactPackageProvider.cs` and make the following changes:
```

docs/native-modules-csharp-codegen.md
```diff
@@ -0,0 +1,94 @@
...
+1. Replace the call to the generated implementation with a reflection based call.
+   ```diff
+            public void CreatePackage(IReactPackageBuilder packageBuilder)
+            {
+   -            CreatePackageImplementation(packageBuilder);
+   +            packageBuilder.AddReflectionReactPackageProvider<ReactPackageProvider>();
+            }
+   ```
+
+> Disclaimer: The reflection logic might deprecated in the near future.
```
NickGerleman, line 45:
```suggestion
> Disclaimer: The reflection logic may be removed in the near future.
```

docs/native-modules-csharp-codegen.md
```diff
@@ -0,0 +1,94 @@
...
+
+```diff
+   <PropertyGroup>
+     ...
++    <ReactNativeCodeGenEnabled>true</ReactNativeCodeGenEnabled>
+   </PropertyGourp>
+```
+Next you will have to add the reflection based generation. Open `ReactPackageProvider.cs` and make the folowing changes:
+
+1. Add a using statment at the top fo the file
```
asklar, line 30:
```suggestion
1. Add a using statement at the top of the file
```

docs/native-modules-csharp-codegen.md
asklar, line 45:
```suggestion
> Disclaimer: The reflection logic might be deprecated in the near future.
```

docs/native-modules-csharp-codegen.md
```diff
@@ -0,0 +1,94 @@
...
+   ```diff
+     using Microsoft.ReactNative;
+   + using Microsoft.ReactNative.Managed;
+   ```
+
+1. Replace the call to the generated implementation with a reflection based call.
```
asklar, line 36:
```suggestion
1. Replace the call to the generated implementation with a reflection-based call.
```

docs/native-modules-csharp-codegen.md
```diff
@@ -0,0 +1,94 @@
...
+   -            CreatePackageImplementation(packageBuilder);
+   +            packageBuilder.AddReflectionReactPackageProvider<ReactPackageProvider>();
+            }
+   ```
+
+> Disclaimer: The reflection logic might deprecated in the near future.
+
+## Upgrading old projects
+When you upgrade an existing 0.62 project to 0.63 you will already have imports for the shared files so you will get the codegen by default.
+It might be the case that your C# code is not yet set up for this. To enable it you'll need to do the following:
```
asklar, line 49:
```suggestion
It might be the case that your C# code is not yet set up for this. To enable codegen you'll need to do the following:
```

docs/native-modules-csharp-codegen.md
```diff
@@ -0,0 +1,94 @@
...
+           }
+   
+           /// <summary>
+           /// This method is implemented by the C# code generator
+           /// </summary>
+           partial void CreatePackageImplementation(IReactPackageBuilder packageBuilder);
+       }
+   }
+   ```
+2. Hookup the regirstation in the constructor App.xaml.cs:
```
asklar, line 71:
```suggestion
2. Hook up the registration in the constructor in `App.xaml.cs`:
```

docs/native-modules-csharp-codegen.md
```diff
@@ -0,0 +1,94 @@
...
+           public App()
+           {
+              ...
+
+              PackageProviders.Add(new Microsoft.ReactNative.Managed.ReactPackageProvider());
+   +          PackageProviders.Add(new ReactPackageProvider());
+   ```
+
+# How it works
+During the build process a new tool is hooked up in MsBuild. It runs at the same time as the code generaton for Xaml happens. The target used is called `ReactNativeManagedCodeGen`. The build writes the arguments to the tool in a response file in the obj folder, for example: `windows\MyProject\obj\x64\Debug\MyProject.ReactNativeCodeGen\MyProject.ReactNativeCodeGen.rsp`. The tool gets takes the same arguments that the C# compiler receives, the source files, the resolved package and project references and the defines. It then uses those to create a workspace using the C# compilers api's ([Roslyn](https://github.com/dotnet/roslyn)) analyzes the code and emits the registrations calls in a generated C# file next to the .rps file. For example `windows\MyProject\obj\x64\Debug\MyProject.ReactNativeCodeGen\MyProject.ReactNativeCodeGen.g.cs`. This C# is then included as part of the build to make it in your application.
```
asklar, line 82:
```suggestion
During the build process a new tool is hooked up in MSBuild. It runs at the same time as the code generation for XAML. The target used is called `ReactNativeManagedCodeGen`. The build writes the arguments to the tool in a response file in the `obj` folder, for example: `windows\MyProject\obj\x64\Debug\MyProject.ReactNativeCodeGen\MyProject.ReactNativeCodeGen.rsp`. The tool takes the same arguments that the C# compiler receives, the source files, the resolved package and project references and the defines. It then uses those to create a workspace using the C# compiler APIs ([Roslyn](https://github.com/dotnet/roslyn)) which analyze the code and emit the registration calls to a generated C# file next to the ` .rsp`  file. For example `windows\MyProject\obj\x64\Debug\MyProject.ReactNativeCodeGen\MyProject.ReactNativeCodeGen.g.cs`. This auto-generated C# file is then included into your application.
```

docs/native-modules-csharp-codegen.md
```diff
@@ -0,0 +1,94 @@
...
+
+Your application does need to have two pieces of code to ensure it is hooked up. If you use the templates to create the project, there is nothing for you to do as it is all take care of. This paragraph just intends to explain how it works
```
asklar, line 84:
```suggestion
Your application needs to have two pieces of code to ensure it is hooked up. If you use the templates to create the project, everything is taken care of. This paragraph just intends to explain how it works.
```

docs/native-modules-csharp-codegen.md
```diff
@@ -0,0 +1,94 @@
...
+First it needs to have a partial class called `ReactPackageProvider` in the default namespace of your project. This class must have a partial method declared with the followign signature:
```
asklar, line 85:
```suggestion
First it needs to have a partial class called `ReactPackageProvider` in the default namespace of your project. This class must have a partial method declared with the following signature:
```

docs/native-modules-csharp-codegen.md
```diff
@@ -0,0 +1,94 @@
...
+```c#
+partial void CreatePackageImplementation(IReactPackageBuilder packageBuilder);
+``` 
+which the generated file implements. This class should implement `IReactPackageProvider` and call the partial method from the `CreatePackage` function. 
+You then need to register this class in your App's xaml startup code by adding the following call to the constructor:
+```c#
+PackageProviders.Add(new ReactPackageProvider());
+```
+You can fine the full code if you create a new template, or in the instructions for upgrading old projects above.
```
asklar, line 94:
```suggestion
You can find the full code if you create a new template, or in the instructions for upgrading old projects above.
```

website/sidebars.json
```diff
@@ -18,7 +18,8 @@
       "native-modules-marshalling-data",
       "native-modules-async",
       "native-modules-jsvalue",
-      "native-modules-advanced"
+      "native-modules-advanced",
+      "native-modules-nuget"
```
asklar, line 22: I think this is supposed to list native-modules-csharp-codegen instead of nuget?
```suggestion
      "native-modules-csharp-codegen"
```

docs/native-modules-csharp-codegen.md
```diff
@@ -0,0 +1,94 @@
+---
+id: native-modules-nuget
```
asklar, line 2:
```suggestion
id: native-modules-csharp-codegen
```

PR #224, opened 2020-08-31
Update supported-community-modules.md

safaiyeh:
react-native-navigation is the name of another navigation library, using the correct name


NickGerleman:
Thanks for catching this!

PR #230, opened 2020-08-31
Updating macOS getting started guide

harinikmsft:
Making 2 changes based on versioning discussion:
- Updated the Getting started guide for macOS for v0.62 with instructions for new apps as well as upgrading existing apps.
- Updated the "next" version to indicate v0.63 is not supported in React Native for macOS yet and point to the 0.62 version as latest. The link in this file will be broken until website version is upgraded.


alloy (approved):
✨

PR #232, opened 2020-08-31
Update rnm-getting-started.md

asklar:
We don't have a 62 getting started for RNM

We will likely need to update the versioned docs for 62 in order to have a 62 doc for RNM, @jonthysell is that right?

### DIFF
--- a/docs/rnm-getting-started.md
+++ b/docs/rnm-getting-started.md
@@ -6,7 +6,7 @@ title: Get Started with macOS
 This guide will help you get started on setting up your very first React Native for macOS app.
 
 >** Latest stable version available for React Native for macOS is 0.62**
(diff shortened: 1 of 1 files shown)


harinikmsft:
I made that link opportunistically - once the website version is updated, that link should work correctly. We do not want to point to the "docs" version because that will point to 0.63 version

asklar:
> I made that link opportunistically - once the website version is updated, that link should work correctly. We do not want to point to the "docs" version because that will point to 0.63 version
@harinikmsft That broke the docs CI though because that page doesn't exist today yet. Once the docs for 0.63 are snapshotted, we can fix this link to point to 62. Does that make sense?

harinikmsft:
I see, ok. I can update the link after the fact later. Thanks.

Issue #235, opened 2020-09-02
Inconsistent Syntax Highlighting For Commands


NickGerleman:
We see a bunch of examples for commands that get colors at relatively arbitrary points. Seems like we're trying to interpret the Snippet as a programming language instead of a command or raw text?

 E.g. 
![image](https://user-images.githubusercontent.com/835219/92002666-f4b68880-ecf4-11ea-8692-273cc877fc49.png)

jonthysell:
When you use the ``` markdown tags, you should specify the language so the correct highlighting is applied (the auto-detect isn't perfect). Use `cmd` for command-line commands, ie:

> \```cmd
> echo Hello World
>  \```

outputs 

```cmd
echo Hello World
```

Using

> \```
> echo Hello World
>  \```

outputs

```
echo Hello World
```

chrisglein:
Okay, sounds like we just need to scrub the docs and add language specifiers in the right places. Tagged as "good first issue".

chrisglein:
@chiaramooney Have you covered all the pages at this point? If your done passing over these, with your merged PR I think you can close this.

chiaramooney:
All pages covered.

PR #241, opened 2020-09-03
Fix Inconsistent Syntax (#235)

chiaramooney:


docs/getting-started.md
```diff
@@ -24,22 +24,22 @@ npx react-native init <projectName> --template react-native@^0.63.2
...
-```
+```cmd
 npx react-native-windows-init --overwrite
 ```
 
 > The --overwrite flag is a temporary measure that ensures the correct files are copied to metro.config.js for the metro bundler to work with Windows. If you are starting a new app, this should have no impact. If you are adding Windows to your existing app and you have modified the metro.config.js file, please back up your changes, run the command and copy over to take effect. We are tracking [this issue here](https://github.com/microsoft/react-native-windows/issues/4698).
 
 Here are the options that `react-native-windows-init` takes:
-```none
+```bash
```
asklar, line 42: this should probably stay as none (it's just text output from running the script with `--help`), or better yet convert it into a markdown table instead :)

docs/getting-started.md
```diff
@@ -24,22 +24,22 @@ npx react-native init <projectName> --template react-native@^0.63.2
 
 Once your project has been initialized, React Native will have created a new sub directory where all your generated files live.
 
-```
+```cmd
```
asklar, line 27: nice! :)

We should probably change the `<projectName>` to `projectName` so that folks can copy and paste the command directly without risk of missing deleting the `<>` which would unintendedly redirect input/output to a file in `cmd`

docs/getting-started.md
chiaramooney: The website seems to want to highlight random sections of the table with ```none. I'll attach a photo below.  Maybe I'll just go ahead and convert it to a table then.
Capture2

asklar: Ah yeah, that looks wonky, thanks for fixing it @chiaramooney !
FYI here is the list of syntaxes/languages that GitHub understands:
https://github.com/github/linguist/blob/master/vendor/README.md

Note that `cmd` isn't there sadly but powershell is so it should be ok to use instead for simple commands like `cd`

chrisglein: > FYI here is the list of syntaxes/languages that GitHub understands:
> https://github.com/github/linguist/blob/master/vendor/README.md

Do we know what the RN website style supports? Is it based on the same core as GitHub's markdown rendering, or something else?

asklar: @chrisglein So it's actually more complicated than just GitHub's markdown (of course...). We use Docusaurus to produce the content from the markdown files, which I believe RN website does too, so the list I posted earlier might not be 100% a match to Docusaurus.

GitHub does have cmd support but it isn't called cmd which makes it hard to find so I originally missed it. It's called Batchfile and it has a few aliases you can use (instead of cmd): `bat`, `batch`, `dosbatch`, `winbatch`.

https://github.com/github/linguist/blob/b410c72001e5023f9fdc9f0d601713878a167c90/lib/linguist/languages.yml#L439

The best way for @chiaramooney to test this is to clone the website repo, make the changes locally, run `yarn build` and `yarn start` locally then navigate to the docs (under /docs/next) and see how it is rendered.

chiaramooney: Sounds good I'll try this today.

website/versioned_docs/version-0.63/getting-started.md
```diff
@@ -40,17 +40,15 @@ npx react-native-windows-init --overwrite
...
-                                [string] [choices: "cs", "cpp"] [default: "cpp"]
-  --overwrite  Overwrite any existing files without prompting          [boolean]
-  --useWinUI3  Targets WinUI 3.0 (Preview) instead of UWP XAML         [boolean]
-  ```
+| Option      | Input Type                            | Description                                      |
+|-------------|---------------------------------------|--------------------------------------------------|
+| --help      | boolean                               | Show help.                                       |
+| --version   | string                                | The version of react-native-windows to use.      |
+| --namespace | string                                | The native project namespace.                    |
+| --verbose   | string                                | Enables logging.                                 |
```
chrisglein, line 48:
```suggestion
| --verbose   | boolean                               | Enables logging.                                 |
```

chrisglein (changes requested):

docs/getting-started.md
```diff
@@ -24,32 +24,30 @@ npx react-native init <projectName> --template react-native@^0.63.2
...
-  ```
+| Option      | Input Type                           | Description                                      |
+|-------------|--------------------------------------|--------------------------------------------------|
+| --help      | boolean                              | Show help.                                       |
+| --version   | string                               | The version of react-native-windows to use.      |
+| --namespace | string                               | The native project namespace.                    | 
+| --verbose   | boolean                              | Enables logging.                                 |
+| --language  | string ["cs","cpp"] [default: "cpp"] | Which language the app is written in.            |
+| --overwrite | boolean                              | Overwrite any existing files without prompting.  |
+| --useWinUI3 | boolean                              | Targets WinUI 3.0 (Preview) instead of UWP XAML. |
```
asklar, line 50:
```suggestion
| Option      | Input Type                           | Description                                      |
|-------------|--------------------------------------|--------------------------------------------------|
| `--help`      | boolean                              | Show help.                                       |
| `--version`  | string                               | The version of react-native-windows to use.      |
| `--namespace` | string                               | The native project namespace.                    | 
| `--verbose`   | boolean                              | Enables logging.                                 |
| `--language`  | string ["cs","cpp"] [default: "cpp"] | Which language the app is written in.            |
| `--overwrite` | boolean                              | Overwrite any existing files without prompting.  |
| `--useWinUI3` | boolean                              | Targets WinUI 3.0 (Preview) instead of UWP XAML. |
```

asklar (approved):
consider code-escaping the options in the table to make them stand out, otherwise looks good :)

chrisglein (approved):

PR #242, opened 2020-09-04
Add C# - C++ interop instructions

asklar:
Add C#-C++ interop instructions since we've had a number of people get confused/blocked by the indecipherable .net error
`0x80131040 : The located assembly's manifest definition does not match the assembly reference.`

Fixes https://github.com/microsoft/react-native-windows/issues/5858


docs/native-modules.md
```diff
@@ -470,3 +470,30 @@ For events, you'll see that we created an instance of `NativeEventEmitter` passi
 Although React Native for Windows supports writing modules in both C# and C++, you should be aware that your choice of language could impact performance of apps that consume your module. Modules written in C# rely on the CLR. At app launch, if there are _any_ C# dependencies, the app will load the CLR which will increase the launch time for the application. Note that this is a one-time cost regardless of the number of C# dependencies that your app relies on.
 
 That said, we recognize the engineering efficiency that comes with writing a module in C#. We strive to maintain parity in developer experience between C# and C++. If your app or module already uses C# (perhaps because it is migrating from the React Native for Windows legacy platform), you should feel empowered to continue to use C#. That said, modules that Microsoft contributes to will be written in C++ to ensure the highest level of performance. 
+
+<div style="padding: 10px; background: #ee2020;color: #fff">
```
chrisglein, line 474: Do we have a better option for blocking these styles in markdown?

chrisglein (approved):

docs/native-modules.md
asklar: We can add a link to a stylesheet to siteconfig.js but it has to be hosted somewhere already, and I don't see a way to publish a css file from docusaurus v1

PR #243, opened 2020-09-04
Native modules doc updates

asklar:
Update native docs with corrected information and add troubleshooting and info about methods returning complex types like string.

Fixes: https://github.com/microsoft/react-native-windows/issues/5910

### DIFF
--- a/docs/native-modules-setup.md
+++ b/docs/native-modules-setup.md
@@ -205,6 +205,26 @@ To make sure that everything is working, you'll want to try building `MyLibrary`
 
 You have now created the scaffolding to build a native module or view manager. Now it's time to add the business logic to the module - follow the steps described in the [Native Modules](native-modules.md) and [View Managers](view-managers.md) documents.
 
+### Making your module ready for consumption in an app
+
+You will need to edit your project file manually to touch up the paths that it uses to reference project references and NuGet packages.
+1. When you add a reference to the `Microsoft.ReactNative.Cxx` project in VS, it will use a path like `..\..\node_modules\react-native-windows\Microsoft.ReactNative.Cxx\Microsoft.ReactNative.Cxx.vcxproj`. This however isn't going to work for a different app. 
+
+Instead you will want to replace that with a reference to the react-native-windows project that the app project might be using:
+```xml
+  <PropertyGroup>
+    <ReactNativeWindowsDir Condition="'$(ReactNativeWindowsDir)' == ''">$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), 'node_modules\react-native-windows\package.json'))\node_modules\react-native-windows\</ReactNativeWindowsDir>
+  </PropertyGroup>
+  <ImportGroup Label="Shared">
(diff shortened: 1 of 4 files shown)


chrisglein (approved):

docs/native-modules-setup.md
```diff
@@ -58,11 +58,11 @@ At this point, follow the steps below to add Windows support to the newly create
...
 ### Updating your package.json
 
-You'll need to ensure you have version 0.62 of both `react-native` and `react-native-windows`. In the directory for your native module project, you can update the dependencies with the following:
+You'll need to ensure you have version 0.63 of both `react-native` and `react-native-windows`. In the directory for your native module project, you can update the dependencies with the following:
 
 ```cmd
-yarn add react-native@0.62 --dev
-yarn add react-native-windows@0.62 --peer
+yarn add react-native@0.63 --dev
+yarn add react-native-windows@0.63 --peer
```
stmoy, line 65: I thought @acoates-ms said something about adding a peer dependency locking the module into a particular version of RNW, meaning it would break as the versions rev. Is this true?

stmoy (approved):
Thank you for these updates!

docs/native-modules-setup.md
acoates-ms: Yeah, these values should be dependent on what the min version of RNW you want to support.   So this should be something like:
`yarn add react-native-windows@>=0.62 --peer`

Also, if you are working on an existing community project, you can't just update the RN dev version without making changes to the android/ios test projects assuming they have some.  --  You will likely need to do some kind of RN update work flow, depending on what is in the repo.

docs/native-modules-setup.md
```diff
@@ -222,7 +242,7 @@ If you are working on an existing module that already has iOS and Android sample
 4. Open the solution with Visual Studio and [link native module](native-modules-using.md).
 > The project should build correctly at this point, but we still need to setup some special metro configurations for Windows in order to run the app without breaking iOS and Android bundling.
 
-5. Add `metro.config.windows` for Windows bundling ([example](https://github.com/react-native-community/react-native-webview/blob/master/metro.config.windows.js)).
+5. Add `metro.config.windows` for Windows bundling ([example](https://github.com/react-native-community/react-native-webview/blob/master/metro.config.windows.js)). Make sure the config file is at the root of the repo (see [Metro bug #588](https://github.com/facebook/metro/issues/588)).
```
acoates-ms, line 245: If we have 0.62 be the min version, then we shouldn't need to do anything special for metro.config.  -- I think I'd rather keep the docs as only whats needed to support >=0.62, and remove all the logic about metro.config.

asklar: Unfortuntely the part about not adding metro.config.js logic is not true, as we have windows specific behavior like react-native bundle failing when being invoked from msbuild because msbuild writes to ProjectImports.zip, see #5050.

docs/native-modules-setup.md
asklar: there is a thread on what the scope of validation that we can do is. I'm going to defer creating the guidance for "update/validate other platforms" until that gets settled.

asklar: Filed https://github.com/microsoft/react-native-windows-samples/issues/244

PR #251, opened 2020-09-09
Custom styling CSS

asklar:
Move some custom styling onto CSS

### DIFF
--- a/website/static/css/custom.css
+++ b/website/static/css/custom.css
@@ -1,12 +1,14 @@
 /* Custom styles */
 
-div.warning   {
+.warning   {
     background: #cc2020;
-    padding: 6pt;
+    padding: 4pt;
     border-radius: 6pt;
+    color: #fff;
 }
 
-div.warning > h3, div.warning > h4 {
+div.warning > h3, 
+div.warning > h4   {
     color: #fff;
-    font-weight: lighter;
+    font-weight: normal;
 }
--- a/website/static/css/custom.css
+++ b/website/static/css/custom.css
@@ -1,0 +1,12 @@
+/* Custom styles */
+
+div.warning   {
+    background: #cc2020;
+    padding: 6pt;
+    border-radius: 6pt;
+}
+
+div.warning > h3, div.warning > h4 {
+    color: #fff;
+    font-weight: lighter;
+}
--- a/docs/native-modules.md
+++ b/docs/native-modules.md
@@ -519,9 +519,9 @@ Although React Native for Windows supports writing modules in both C# and C++, y
 
 That said, we recognize the engineering efficiency that comes with writing a module in C#. We strive to maintain parity in developer experience between C# and C++. If your app or module already uses C# (perhaps because it is migrating from the React Native for Windows legacy platform), you should feel empowered to continue to use C#. That said, modules that Microsoft contributes to will be written in C++ to ensure the highest level of performance. 
 
-<div style="padding: 10px; background: #ee2020;color: #fff">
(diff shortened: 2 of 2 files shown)


website/static/css/custom.css
```diff
@@ -0,0 +1,12 @@
...
+div.warning   {
+    background: #cc2020;
+    padding: 6pt;
+    border-radius: 6pt;
+}
+
+div.warning > h3, div.warning > h4 {
+    color: #fff;
+    font-weight: lighter;
+}
```
chrisglein, line 12:
```suggestion
.warning {
    background: #cc2020;
    padding: 6pt;
    border-radius: 6pt;
    color: #fff;
}

.warning > h3,
.warning > h4 {
    font-weight: lighter;
} 
```

- You can put `color` on the parent element and that'll be the default for the parent.
- I'd strip the `div` requirement and just base it on the class (`.warning`).

chrisglein (approved):

website/static/css/custom.css
chrisglein: Boo, there's a color override for the header styles in place:
```css
.mainContainer h1, .mainContainer h2, .mainContainer h3, .mainContainer h4, .mainContainer h5, .mainContainer h6 {
    color: #1a1a1a;
}
```
So yeah, the above won't work without overriding that.

chrisglein:
```suggestion
.warning {
    background: #cc2020;
    padding: 6pt;
    border-radius: 6pt;
    color: #fff;
}

.warning > h3,
.warning > h4 {
    font-weight: lighter;
    color: #fff;
} 
```
By having it in both you cover the case of someone putting text inside the red box without being a header. The other header types are left unaddressed, but /shrug.

asklar:
```suggestion
.warning   {
    background: #cc2020;
    padding: 4pt;
    border-radius: 6pt;
    color: #fff;
}

div.warning > h3, 
div.warning > h4   {
    color: #fff;
    font-weight: normal;
}
```

website/static/css/custom.css
```diff
@@ -0,0 +1,14 @@
...
+
+.warning   {
+    background: #cc2020;
+    padding: 4pt;
+    border-radius: 6pt;
+    color: #fff;
+}
+
+div.warning > h3, 
+div.warning > h4   {
```
chrisglein, line 11:
```suggestion
.warning > h3, 
.warning > h4   {

```

asklar: @chrisglein  that no worky:
image
image

chrisglein: > @chrisglein that no worky:

Ugh, the new style isn't winning because it's earlier in the file and the rules have the same specificity. CSS rules don't just go in order in the file, they're also weighed by specificity. So `.mainContainer h3` is the same level as `.warning > h3` . By making `div.warning > h3` you're increasing the specificity enough to override the file order. Alternate fixes would be either:
- Move the `.warning` CSS rules lower in the file
- Update the rule `.mainContainer .warning h3` so that it is more specific in a way that better matches the overriden rule

asklar: @chrisglein we don't control the order of inclusion as that's done by docusaurus. 
If/we move to docusaurus v2, a lot of this can go away since we can just embed react components instead of dealing with css styling

asklar: filed #269, I've been talking to the docusaurus dev about migrating, we need some features from v1 that were not ported to v2 yet before we can move.

PR #268, opened 2020-09-21
Updated native module setup docs for 61, 62, 63, and master

### LINKED ISSUE #249: Update Native Modules + Visual Studio documentation
While trying to port [react-native-config](https://github.com/luggit/react-native-config) I've followed the new 0.63 module guide, and encountered some issues:

- In [Adding React Native Windows to the Visual Studio Solution](https://microsoft.github.io/react-native-windows/docs/native-modules-setup#adding-react-native-windows-to-the-visual-studio-solution): there are no `Microsoft.ReactNative.SharedManaged\Microsoft.ReactNative.SharedManaged.shproj` nor `ReactWindowsCore\ReactWindowsCore.vcxproj` in the `node_modules\react-native-windows` folder. The module builds and works without those though.
- Adding `Microsoft.ReactNative.Cxx` as a reference in the `Shared Projects` only works after reloading the solution. Before that, no such project is present in that tab.
- First build will fail with missing Folly headers, subsequent builds works.
- When adding the module (by using `npm i ..\the_module`) to a RNW app (created by following the guide) the app will fail to build if `node_modules` folder is present in the module folder.
- Running the RNW app with `npx react-native run-windows` fails with:
```
Failed to deploy:     11>C:\[app folder]\node_modules\react-native-config\windows\RNCConfig\RNCConfig.vcxproj : error MSB4057: The target "Deploy" does not exist in the project.
```
(RNCConfig.vcxproj is the module .vcxproj file)
The RN server will start, and the app when started from Visual Studio will connect to that server and work correctly.

### LINKED ISSUE #253: Update package.json guidance for native modules
https://microsoft.github.io/react-native-windows/docs/native-modules-setup#updating-your-packagejson

```
yarn add react-native@0.63 --dev
yarn add react-native-windows@0.63 --peer
```

Depending on which versions of iOS and Android the module supports, this guidance potentially breaks the module when Windows support is added. 

What is the guidance around adding Windows support to a module without breaking iOS and Android?

jonthysell:
A variety of necessary updates to the native module setup docs.

Closes #249 
Closes #253 
Closes #273


docs/native-modules-setup.md
```diff
@@ -45,154 +45,64 @@ The steps to create a new native module library project are:
...
 
-1. Close the solution with `File` > `Close Solution`.
-1. Rename that top `MyLibrary` directory `windows`.
-1. Re-open the solution file at `windows\MyLibrary.sln`.
+> For more information on how NPM dependencies work, see [Specifying dependencies and devDependencies in a package.json file](https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file).
 
-> *Optional*: Whether you're creating a new native module from scratch, or adding windows support to an existing ios/android module, if you're using git, you'll want to add Visual Studio-specific entries in your project's `.gitignore`.
-> The simplest way to do this is to get [VisualStudio.gitignore](https://raw.githubusercontent.com/github/gitignore/master/VisualStudio.gitignore) and save it as `windows\.gitignore`.
+You can use the `npm info` command to find the correct versions to use. Let's assume you plan on building against the latest `canary` version of `react-native-windows`.
```
NickGerleman, line 63: Might be better to point to `latest` instead of canary for examples, since Canary has weird version numbers, etc.

docs/native-modules-setup.md
```diff
@@ -45,154 +45,64 @@ The steps to create a new native module library project are:
...
 
-After you've added the reference, you need to make sure it doesn't copy itself into your build (otherwise it'll cause build conflicts down the line when you're trying on consume your library):
+Now you should be ready to add Windows support with the CLI.
 
-1. Open the Solution Explorer sidebar.
-1. Under your `MyLibrary` project, expand the `References`.
-1. Right-click on `Microsoft.ReactNative`.
-1. Select `Properties`.
-1. Under `Build`, Change `Copy Local` to `False`.
+## Adding Windows support to an existing library
```
NickGerleman, line 91: Does `create-react-native-module` add example projects for iOS/Android that would need to be revved?

docs/native-modules-setup.md
```diff
@@ -45,154 +45,64 @@ The steps to create a new native module library project are:
...
-1. Right-click on your `MyLibrary` project.
-1. Select `Add` > `Reference`.
-1. Select `Shared Projects` on the left-hand side.
-1. Check the box next to `Microsoft.ReactNative.SharedManaged`.
-1. Click `OK`.
+This defaults to a C++/WinRT project. If you want to create a C# based native module project, use:
 
-If you're writing in C++, you'll want to add `Microsoft.ReactNative.Cxx`:
+```bat
+npx react-native-windows-init --version canary --projectType lib --language cs --overwrite
```
NickGerleman, line 102: What does `react-native-windows-init` lib mode do for dependencies/peer dependencies? Feels like we might need to do something manually here. I.e. we don't want any deps on react-native-windows, but do want a permissive peer dependency.

docs/native-modules-setup.md
```diff
@@ -45,154 +45,64 @@ The steps to create a new native module library project are:
...
 
-1. Open the Solution Explorer sidebar.
-1. Under your `MyLibrary` project, expand the `References`.
-1. Right-click on `Microsoft.ReactNative`.
-1. Select `Properties`.
-1. Under `Build`, Change `Copy Local` to `False`.
+## Adding Windows support to an existing library
 
-Now, you're technically ready to go, but in order to improve the developer experience, it's also **highly recommended** to also add a reference to the appropriate helper shared project. These projects contain the attributes (C#) and macros (C++) as described in the [Native Modules](native-modules.md) and [View Managers](view-managers.md) documents.
+Similar to adding Windows support to an app project, you can use the CLI to add Windows support to a native module project, but you'll need to specify `--projectType lib`:
```
NickGerleman, line 93: I'm not sure if there's a good solution, but I remember confusion before when someone tried running commands in our docs at the root of a monorepo. Not sure if adding something about "run this in the directory with your module package" would be useful or not.

website/versioned_docs/version-0.61/native-modules-setup.md
```diff
@@ -1,232 +1,311 @@
...
+
+For each project, you'll do the following:
+
+1. Open the Solution Explorer sidebar.
+1. Right-click on the `ReactNative` folder.
+1. Select `Add` > `Existing Project...`.
+1. Select the project file and click `Open`.
+
+When you are done, your solution should look like this:
+![native module dependencies](assets/native-module-dependencies61.png)
```
NickGerleman, line 188: I wonder if there's a way we could log recurring work for bits like this that need to be updated for each release. I get the feeling we could otherwise just forget. Part of me thinks it might be better to leave out bits that we think would get quickly outdated.

docs/native-modules-setup.md
jonthysell: Each version of the doc points to the version that doc is for. The files under `docs` are for the "next" release, after `latest`, which right now is `canary`. Then the next preview starts I'll change it to `preview`.

docs/native-modules-setup.md
jonthysell: `create-react-native-module` *can* create example projects with a flag.

I'm not sure what guidance to give about the main project's dependencies, let alone example app dependencies. The instructions here make sure that you have the right files on disk to build. And even if you publish to npm (with or without the changes to package.json), an RNW app will be able to consume your native module.

This is least intrusive way to implement modules without any assumptions or proscriptions about the module's maintenance.

docs/native-modules-setup.md
jonthysell: Th logic is here: https://github.com/microsoft/react-native-windows/blob/master/packages/react-native-windows-init/src/Cli.ts#L331

From what I can see, only apps depend directly on react, react-native, and/or react-native-windows. Libraries may have zero dependencies, and at minimum just have react-native as a dev dependency.

So if I see dependencies with react-native (probably an app), we use the previous logic of inserting rnw into dependencies and running yarn/npm install.

If I see 'react-native' as a dev dependency (probably a lib), then add rnw as a dev dependency with yarn/npm

website/versioned_docs/version-0.61/native-modules-setup.md
jonthysell: I think we need explicit owners of each doc that are responsible for keeping them up to date.

docs/native-modules-setup.md
NickGerleman: Makes sense. I think we need to add a step asking to add a peer dependency then, since the CLI just adds a dev dependency.

docs/native-modules-setup.md
NickGerleman: My concern was more that someone could read this, then rev existing examples versions without necessarily keeping them up to date. I.e. if we create iOS/Android examples for a version, then run these commands, there is some extra work to do still that has tripped people up before.

Could we add a sentence pointing folks towards upgrade-helper if there are a existing sample applications?

docs/native-modules-setup.md
jonthysell: I can make the language clearer that the location to run the command is usually the repo root but not always.

docs/native-modules-setup.md
jonthysell: I think there needs to some "global" statement that if you're adding windows implementation to a community module, you'll need to work with the module owners when making any `package.json` changes.

docs/native-modules-setup.md
jonthysell: BTW, is there any way we can "unpublish" `preview` from npm when there's no active preview? Or re-point `preview` to latest once we officially release? It's a little weird when `preview` predates `latest`.

NickGerleman: We can remove tags if we want to. `react-native` uses `next` instead of `preview`, and kind of does the same thing where `next precedes` `latest` in certain cases (I think there's is honestly more confusing).

docs/native-modules-setup.md
```diff
@@ -45,154 +45,66 @@ The steps to create a new native module library project are:
...
-
-Additionally, for C++/WinRT projects, you'll need to change the following:
-
-1. Right click on the project and choose `Manage NuGet Packages...`
-    1. Select version 2.0.200316.3 for Microsoft.Windows.CppWinRT package.
-1. Right-click on the project and choose `Properties`
-    1. Under `Linker > Windows Metadata` set `Generate Windows Metadata` to `Yes`.
-
-Now, before we go any further we'll want to rename the root directory of the Windows native code to `windows` to match the peer `android` and `ios` directories:
+Unfortunately the default library template targets older versions of `react` and `react-native` that Windows supports, so you'll need to upgrade to newer versions of `react` and `react-native` in order to add supprot for `react-native-windows`.
```
asklar, line 59:
```suggestion
Unfortunately the default library template targets older versions of `react` and `react-native` that Windows supports, so you'll need to upgrade to newer versions of `react` and `react-native` in order to add support for `react-native-windows`.
```

docs/native-modules-setup.md
```diff
@@ -45,154 +45,66 @@ The steps to create a new native module library project are:
...
+Take the result of that command (let's say it's `16.13.1`) and use it to upgrade the dev dependency:
 
-When you are done, your solution should look like this:
-![native module dependencies](assets/native-module-dependencies.png)
+```bat
+yarn upgrade react@16.13.1 --dev
+```
 
-You now have all of the React Native Windows projects to your solution. Next we're going to reference them in our `MyLibrary` project.
+You'll need to repeat the steps for `react-native`, ie:
```
asklar, line 79:
```suggestion
You'll need to repeat the steps for `react-native`, i.e.:
```

docs/native-modules-setup.md
```diff
@@ -45,154 +45,66 @@ The steps to create a new native module library project are:
...
 
-1. Open the Solution Explorer sidebar.
-1. Under your `MyLibrary` project, expand the `References`.
-1. Right-click on `Microsoft.ReactNative`.
-1. Select `Properties`.
-1. Under `Build`, Change `Copy Local` to `False`.
+## Adding Windows support to an existing library
 
-Now, you're technically ready to go, but in order to improve the developer experience, it's also **highly recommended** to also add a reference to the appropriate helper shared project. These projects contain the attributes (C#) and macros (C++) as described in the [Native Modules](native-modules.md) and [View Managers](view-managers.md) documents.
+Similar to adding Windows support to an app project, you can use the CLI to add Windows support to a native module project, but you'll need to specify `--projectType lib`:
```
asklar, line 95:
```suggestion
Similarly to adding Windows support to an app project, you can use the CLI to add Windows support to a native module project, by specifying `--projectType lib`:
```

docs/native-modules-setup.md
```diff
@@ -207,22 +119,22 @@ You have now created the scaffolding to build a native module or view manager. N
...
-    <ReactNativeWindowsDir Condition="'$(ReactNativeWindowsDir)' == ''">$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), 'node_modules\react-native-windows\package.json'))\node_modules\react-native-windows\</ReactNativeWindowsDir>
-  </PropertyGroup>
-  <ImportGroup Label="Shared">
-    <Import Project="$(ReactNativeWindowsDir)\Microsoft.ReactNative.Cxx\Microsoft.ReactNative.Cxx.vcxitems" Label="Shared" />
-  </ImportGroup>
+  <Import Project="..\packages\NuGetPackage.1.0.0.0\build\native\NuGetPackage.props" Condition="Exists('..\packages\NuGetPackage.1.0.0.0\build\native\NuGetPackage.props')" />
 ```
 
-2. NuGet packages will show up in the vcxproj as relative references too, e.g. `..\packages\...`. We need these to use the solution directory instead, so replace all mentions of this form with something like:
+to this:
```
asklar, line 134: consider using the diff format (```diff)

docs/native-modules-setup.md
```diff
@@ -14,17 +16,15 @@ This guide will help set you up with the Visual Studio infrastructure to author
 
 Make sure you have installed all of the [development dependencies](rnw-dependencies.md).
 
-If you're planning on writing in C++, you *must* install the [C++/WinRT Visual Studio Extension](https://marketplace.visualstudio.com/items?itemName=CppWinRTTeam.cppwinrt101804264).
```
asklar, line 17: shouldn't we keep this on? or at least make it a recommendation?

asklar: actually it might make sense for rnw-dependencices to install it

jonthysell: AFAIK this only adds the C++/WinRT project templates to VS. Since the CLI is creating the project for them, they don't really need it.

docs/native-modules-setup.md
jonthysell: We can take this offline.

PR #270, opened 2020-09-22
UWP native code docs

### LINKED ISSUE #264: Create a troubleshooting guide for UWP apps and C#-C++ interop
This is applicable to not only RNW but to all UWP, but there aren't great resources on docs.microsoft.com. 
My proposal is:
- create a UWP/C#/C++ troubleshooting guide
- move the content I wrote about .net native there
- augment it with things like consuming C#-with-XAML from a cppwinrt app: https://github.com/microsoft/react-native-windows/issues/6003
- point to the [cppwinrt troubleshooting guide](https://docs.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/troubleshooting)
- point to instructions about installing/deploying to store

asklar:
Add a doc about RNW apps as UWP apps

Fixes #264

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -39,21 +39,13 @@ npx react-native-windows-init --overwrite
 > The --overwrite flag is a temporary measure that ensures the correct files are copied to metro.config.js for the metro bundler to work with Windows. If you are starting a new app, this should have no impact. If you are adding Windows to your existing app and you have modified the metro.config.js file, please back up your changes, run the command and copy over to take effect. We are tracking [this issue here](https://github.com/microsoft/react-native-windows/issues/4698).
 
 Here are the options that `react-native-windows-init` takes:
-| Option      | Input Type                           | Description                                      |
-|-------------|--------------------------------------|--------------------------------------------------|
-| --help      | boolean                              | Show help.                                       |
-| --version   | string                               | The version of react-native-windows to use.      |
-| --namespace | string                               | The native project namespace.                    | 
-| --verbose   | boolean                              | Enables logging.                                 |
-| --language  | string ["cs","cpp"] [default: "cpp"] | Which language the app is written in.            |
-| --overwrite | boolean                              | Overwrite any existing files without prompting.  |
-| Option      | Input Type                           | Description                                      |
-|-------------|--------------------------------------|--------------------------------------------------|
+| Option      | Input Type                           | Description                                        |
+|-------------|--------------------------------------|----------------------------------------------------|
 | `--help`      | boolean                              | Show help.                                       |
-| `--version`  | string                               | The version of react-native-windows to use.      |
+| `--version`  | string                               | The version of react-native-windows to use.       |
 | `--namespace` | string                               | The native project namespace.                    | 
 | `--verbose`   | boolean                              | Enables logging.                                 |
-| `--language`  | string ["cs","cpp"] [default: "cpp"] | Which language the app is written in.            |
+| `--language`  | string ["`cs`","`cpp`"] [default: "cpp"] | Which language the app is written in.            |
 | `--overwrite` | boolean                              | Overwrite any existing files without prompting.  |
 | `--useWinUI3` | boolean                              | Targets WinUI 3.0 (Preview) instead of UWP XAML. |
 
@@ -110,7 +102,7 @@ Follow these steps to build a version of your app that you can install or publis
 - Open the solution in Visual Studio
 - Select the Release configuration from the Configuration Manager dropdown.
 - Build the solution. You can now launch without first launching Metro.
-- If you want to build an appx package to share or publish, use the Project => Publish => Create App Packages... option.
+- If you want to build an appx package to share or publish, use the **Project** > **Publish** > **Create App Packages...** option.
 
 See also this article for additional details: https://techcommunity.microsoft.com/t5/windows-dev-appconsult/getting-started-with-react-native-for-windows/ba-p/912093#
 </body>
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -45,7 +45,8 @@ Here are the options that `react-native-windows-init` takes:
 | `--version`  | string                               | The version of react-native-windows to use.       |
 | `--namespace` | string                               | The native project namespace.                    | 
 | `--verbose`   | boolean                              | Enables logging.                                 |
-| `--language`  | string ["`cs`","`cpp`"] [default: "cpp"] | Which language the app is written in.            |
+| `--language`  | string ["`cs`","`cpp`"] [default: "`cpp`"] | Which language the app is written in.            |
+
 | `--overwrite` | boolean                              | Overwrite any existing files without prompting.  |
 | `--useWinUI3` | boolean                              | Targets WinUI 3.0 (Preview) instead of UWP XAML. |
 
--- a/docs/native-code.md
+++ b/docs/native-code.md
@@ -18,7 +18,7 @@ The current publicly supported version of .NET has built-in support for WinRT ([
 
 Whether you are new to Windows development, or you are a Win32 desktop app veteran, the following FAQs should answer some common questions.
 
-When you add Windows support to a react native app via the steps described in the [Install the windows extension](getting-started#install-the-windows-extension) section, the react-native-windows CLI will create a UWP app for you. 
+When you add Windows support to a react native app via the steps described in the [Install the windows extension](getting-started.md#install-the-windows-extension) section, the react-native-windows CLI will create a UWP app for you. 
 
 Therefore, <u>**RNW apps are UWP apps**</u> and therefore have the following characteristics:
 - The set of APIs these app can access are a subset of all Windows APIs (i.e. those accessible via WinRT). See:
--- a/docs/native-code.md
+++ b/docs/native-code.md
@@ -37,7 +37,8 @@ React Native Windows apps are signed and packaged. [Packaging](https://docs.micr
 
 
 ### Distribution
-They can be distributed, installed and updated in the following ways:
+React Native Windows apps can be distributed, installed and updated in the following ways:
+
   - via [the Microsoft Store](https://docs.microsoft.com/en-us/windows/apps/desktop/modernize/desktop-to-uwp-distribute).
   - via [your private Store](https://docs.microsoft.com/en-us/microsoft-store/distribute-apps-to-your-employees-microsoft-store-for-business) if you are a business or educational organization. See also [Distribute LOB apps to enterprises](https://docs.microsoft.com/en-us/windows/uwp/publish/distribute-lob-apps-to-enterprises).
   - using [App Installer](https://docs.microsoft.com/en-us/windows/msix/app-installer/installing-windows10-apps-web).
--- a/docs/native-code.md
+++ b/docs/native-code.md
@@ -33,7 +33,8 @@ The app runs inside of an app container - a type of sandbox. This provides apps 
 
 ### Packaging
 
-They are signed and packaged. [Packaging](https://docs.microsoft.com/en-us/windows/uwp/packaging/) is a mechanism through which an app and its dependencies acquire an identity, which is used to determine whether API calls that require system capabilities (e.g. filesystem access) should succeed or not.
+React Native Windows apps are signed and packaged. [Packaging](https://docs.microsoft.com/en-us/windows/uwp/packaging/) is a mechanism through which an app and its dependencies acquire an identity, which is used to determine whether API calls that require system capabilities (e.g. filesystem access) should succeed or not.
+
 
 ### Distribution
 They can be distributed, installed and updated in the following ways:
--- a/docs/native-code.md
+++ b/docs/native-code.md
@@ -5,12 +5,12 @@ title: Working with native code on Windows
 
 ## What is a React Native for Windows app?
 
-A RNW app is a [Universal Windows Platform app](https://docs.microsoft.com/en-us/windows/uwp/get-started/universal-application-platform-guide) (aka UWP or Windows 10 app).
+A RNW app is a [Universal Windows Platform app](https://docs.microsoft.com/windows/uwp/get-started/universal-application-platform-guide) (aka UWP or Windows 10 app).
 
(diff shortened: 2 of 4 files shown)


asklar:
@stevewhims FYI

docs/native-code.md
```diff
@@ -0,0 +1,122 @@
...
+- [Alternatives to Windows APIs in Universal Windows Platform (UWP) apps](https://docs.microsoft.com/en-us/uwp/win32-and-com/alternatives-to-windows-apis-uwp)
+
+### Isolation
+The app runs inside of an app container - a type of sandbox. This provides apps with a secure way to install, access system resources like the filesystem, and lets the system manage their lifetime (e.g. suspending the app when it isn't on the foreground). This means that by default an RNW app cannot access arbitrary filesystem locations, start arbitrary processes, etc. UWP apps that need to access these kinds of capabilities may be able to do so via [App capability declarations](https://docs.microsoft.com/en-us/windows/uwp/packaging/app-capability-declarations).
+
+### Packaging
+
+They are signed and packaged. [Packaging](https://docs.microsoft.com/en-us/windows/uwp/packaging/) is a mechanism through which an app and its dependencies acquire an identity, which is used to determine whether API calls that require system capabilities (e.g. filesystem access) should succeed or not.
```
stmoy, line 36:
```suggestion
React Native Windows apps are signed and packaged. [Packaging](https://docs.microsoft.com/en-us/windows/uwp/packaging/) is a mechanism through which an app and its dependencies acquire an identity, which is used to determine whether API calls that require system capabilities (e.g. filesystem access) should succeed or not.

```

docs/native-code.md
```diff
@@ -0,0 +1,122 @@
...
+
+### Distribution
+They can be distributed, installed and updated in the following ways:
```
stmoy, line 39:
```suggestion
React Native Windows apps can be distributed, installed and updated in the following ways:

```

docs/native-code.md
```diff
@@ -0,0 +1,122 @@
+---
+id: native-code
+title: Working with native code on Windows
+---
+
+## What is a React Native for Windows app?
+
+A RNW app is a [Universal Windows Platform app](https://docs.microsoft.com/en-us/windows/uwp/get-started/universal-application-platform-guide) (aka UWP or Windows 10 app).
```
stmoy, line 8:
```suggestion
A RNW app is a [Universal Windows Platform app](https://docs.microsoft.com/windows/uwp/get-started/universal-application-platform-guide) (aka UWP or Windows 10 app).

```

Throughout, please remove the language tokens from the URLs so that folks visiting from other locales get the right version of the sites.

docs/native-code.md
```diff
@@ -0,0 +1,122 @@
...
+---
+
+## What is a React Native for Windows app?
+
+A RNW app is a [Universal Windows Platform app](https://docs.microsoft.com/en-us/windows/uwp/get-started/universal-application-platform-guide) (aka UWP or Windows 10 app).
+
+The Universal Windows Platform allows you to access a set of common functionality on all Windows devices via the [Windows Runtime](https://docs.microsoft.com/en-us/windows/uwp/winrt-components/) (WinRT). WinRT APIs can be accessed from C++ (via [C++/WinRT](https://docs.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/)), or via .NET C#. 
+
+#### WinRT support in .NET
+The current publicly supported version of .NET has built-in support for WinRT ([.NET UWP](https://docs.microsoft.com/en-us/windows/uwp/winrt-components/creating-windows-runtime-components-in-csharp-and-visual-basic)) whereas future versions of .NET Core starting with .NET Core 5 will only be able to access WinRT APIs via [C#/WinRT](https://docs.microsoft.com/en-us/windows/uwp/csharp-winrt/) in a way similar to what the C++/WinRT projection enables today.
```
stmoy, line 13:
```suggestion
The current publicly supported version of .NET has built-in support for WinRT ([.NET UWP](https://docs.microsoft.com/en-us/windows/uwp/winrt-components/creating-windows-runtime-components-in-csharp-and-visual-basic)). 

Future versions of .NET Core starting with .NET Core 5 will only be able to access WinRT APIs via [C#/WinRT](https://docs.microsoft.com/en-us/windows/uwp/csharp-winrt/) in a way similar to what the C++/WinRT projection enables today.
```

Should we discuss future versions of .NET Core in this doc? I don't understand what the developer action is as a result of this sentence. Is it that their native modules must be written in CsWinRT?

docs/native-code.md
```diff
@@ -0,0 +1,122 @@
...
+### Packaging
+
+They are signed and packaged. [Packaging](https://docs.microsoft.com/en-us/windows/uwp/packaging/) is a mechanism through which an app and its dependencies acquire an identity, which is used to determine whether API calls that require system capabilities (e.g. filesystem access) should succeed or not.
+
+### Distribution
+They can be distributed, installed and updated in the following ways:
+  - via [the Microsoft Store](https://docs.microsoft.com/en-us/windows/apps/desktop/modernize/desktop-to-uwp-distribute).
+  - via [your private Store](https://docs.microsoft.com/en-us/microsoft-store/distribute-apps-to-your-employees-microsoft-store-for-business) if you are a business or educational organization. See also [Distribute LOB apps to enterprises](https://docs.microsoft.com/en-us/windows/uwp/publish/distribute-lob-apps-to-enterprises).
+  - using [App Installer](https://docs.microsoft.com/en-us/windows/msix/app-installer/installing-windows10-apps-web).
+  - It's worth noting that you cannot just "_copy an exe_" as the app package contains more than just the main executable, including an app manifest, assets, dependent framework libraries, etc.
```
stmoy, line 43: This bullet and later bullets don't follow the same pattern as the first three; should this be a separate list? It seems like these bullets are more FYI whereas the first bullets are about how the apps can be distributed.

docs/native-code.md
```diff
@@ -0,0 +1,122 @@
...
+- UWP apps written in C# or other managed languages submitted to the store must be built using the [.NET Native toolchain](https://docs.microsoft.com/en-us/dotnet/framework/net-native/). This is the default when building C# apps in Release mode, **but not in Debug**, so apps built in Debug will be rejected by the Store.
+
+
+### Use of non-WinRT libraries
+Any libraries you use should be built as WinRT components. In other words, you cannot easily link libraries built for win32 desktop apps without additional work.
+  - C++/CX is a dialect of C++ that allows writing UWP apps, however this is **not supported** for writing a RNW app. The article [How to use existing C++ code in a Universal Windows Platform app](https://docs.microsoft.com/en-us/cpp/porting/how-to-use-existing-cpp-code-in-a-universal-windows-platform-app) talks about how to consume non-WinRT libraries in a WinRT context using C++/CX, but most of the content should be applicable to using C++/WinRT which is the supported way to write RNW apps.
```
stmoy, line 50: I'm confused about this bullet as it relates to the native modules docs, which I thought could be written in "pure" C++. 

Meta: I'm not sure what the relationship between these docs and the native modules docs all-up, especially this section.

docs/getting-started.md
```diff
@@ -39,21 +39,13 @@ npx react-native-windows-init --overwrite
...
-|-------------|--------------------------------------|--------------------------------------------------|
+| Option      | Input Type                           | Description                                        |
+|-------------|--------------------------------------|----------------------------------------------------|
 | `--help`      | boolean                              | Show help.                                       |
-| `--version`  | string                               | The version of react-native-windows to use.      |
+| `--version`  | string                               | The version of react-native-windows to use.       |
 | `--namespace` | string                               | The native project namespace.                    | 
 | `--verbose`   | boolean                              | Enables logging.                                 |
-| `--language`  | string ["cs","cpp"] [default: "cpp"] | Which language the app is written in.            |
+| `--language`  | string ["`cs`","`cpp`"] [default: "cpp"] | Which language the app is written in.            |
```
chrisglein, line 48: Why the double quotes?

chrisglein (approved):

docs/native-code.md
```diff
@@ -0,0 +1,122 @@
...
+  - Libraries built for .NET desktop framework cannot be directly accessed by UWP. You can create a .NET Standard library that calls into the .NET framework one, and call from the UWP app into the .NET Standard middleware.
+
+### Local testing and inner loop
+For internal development, you can deploy your app for test purposes by side-loading and deploying via [loose-file registration](https://docs.microsoft.com/en-us/windows/uwp/debug-test-perf/loose-file-registration). When building in **Debug** mode (which is the default), `npx react-native run-windows` performs loose-file registration of your app in order to install it locally. When running `npx react-native run-windows` with the `--release` switch, the CLI will install the real package onto your local machine. This requires the app to be signed and for the certificate it uses for signing to be trusted by the machine the app is going to be installed on. See [Create a certificate for package signing](https://docs.microsoft.com/en-us/windows/msix/package/create-certificate-package-signing) and [Intro to certificates](https://docs.microsoft.com/en-us/windows/uwp/security/certificates).
+
+
+## C# vs. C++ for Native Modules
```
stmoy, line 58: It looks like this was cut-pasted from https://microsoft.github.io/react-native-windows/docs/native-modules#c-vs-c-for-native-modules  but some content is lost (esp. around the CLR, the specific performance implications, etc.) - where did that content go?

stmoy (changes requested):

docs/getting-started.md
asklar: the backquotes styles this as `cs` instead of cs

docs/native-code.md
asklar: .NET Core 5 is about to go live (they already published RC 2). Once they go live, UWP apps written in managed languages should likely be written in .NET Core 5 as .NET UWP will be legacy/deprecated.

docs/getting-started.md
chrisglein: Oh, right. My brain forgot about markdown for a moment and read this is some bizarre string escaping option.

chrisglein:
```suggestion
| `--language`  | string ["`cs`","`cpp`"] [default: "`cpp`"] | Which language the app is written in.            |

```

docs/native-code.md
asklar: Native modules allow you to access the underlying platform, which for UWP you would do via cppwinrt. Yes, technically you can write a pure C++ module without calling any winrt components (e.g. the sample add x+y native module), but that's not what people will usually do.

The point of this paragraph is about "bring some library from nuget". We've seen people ask this on the repo, "can I use this win32 library in the context of a native module?", or they try it and it doesn't work. The info here is to warn them about it and provide some paths forward.

docs/native-code.md
asklar: I deleted the content about loading the CLR since that is not true for UWP apps as they are cross compiled to native.

docs/native-code.md
```diff
@@ -0,0 +1,127 @@
...
+### Distribution
+React Native Windows apps can be distributed, installed and updated in the following ways:
+
+  - via [the Microsoft Store](https://docs.microsoft.com/windows/apps/desktop/modernize/desktop-to-uwp-distribute).
+  - via [your private Store](https://docs.microsoft.com/microsoft-store/distribute-apps-to-your-employees-microsoft-store-for-business) if you are a business or educational organization. See also [Distribute LOB apps to enterprises](https://docs.microsoft.com/windows/uwp/publish/distribute-lob-apps-to-enterprises).
+  - using [App Installer](https://docs.microsoft.com/windows/msix/app-installer/installing-windows10-apps-web).
+
+It's worth noting that you cannot just "_copy an exe_" as the app package contains more than just the main executable, including an app manifest, assets, dependent framework libraries, etc.
+
+In addition, the Store submission process has this requirements:
```
stmoy, line 48:
```suggestion
In addition, the Store submission process has these requirements:

```

docs/native-code.md
stmoy: That's fine - we should still articulate how the choice of language impacts performance or remove this line.

docs/native-code.md
stmoy: I still don't understand what the impact is for RNW developers. Will they need to do something special when .NET Core 5 goes live? Or will that be handled by the RNW platform? Will they need to update their native modules?

asklar: Hmm ok. There still is no official guidance re: .net core 5 and UWP (e.g. https://github.com/microsoft/ProjectReunion/issues/105), it seems like .net core 5 might not support UWP apps "as we know them today" out of the box, so maybe we should have a discussion about what this means separate from this PR.

docs/native-code.md
```diff
@@ -0,0 +1,127 @@
...
+
+C# development brings engineering efficiencies to writing a module or an app. We strive to maintain parity in developer experiences between C# and C++. If your app or module already uses C#, you should feel empowered to continue to use C#. That said, modules that Microsoft contributes to will be written in C++ to ensure the highest level of performance and interoperability.
+
+<div class="warning">
+  <h3>Important</h3>
+  <h4>Mixing C# and C++</h4>
+</div>
+
+### App crashes when trying to load the C# component or instantiate one of its types
+**Error 0x80131040 “The located assembly’s manifest definition does not match the assembly reference”**
```
NickGerleman, line 75: I believe the 0.63+ C++ sample template does some of this by default now.

docs/native-code.md
```diff
@@ -0,0 +1,127 @@
...
+   <ConsumeCSharpModules>true</ConsumeCSharpModules>
+   ```
+
+## Troubleshooting and debugging C++ native modules
+
+So you added a new native module or a new method to a module but it isn't working, **now what?!**
+
+If your method isn't being hit in the VS debugger, something is blocking the call due to a mismatch, likely between the expected and actual types that your method takes/returns.
+
+To debug into what is rejecting the call, set a breakpoint in `CxxNativeModule::invoke` (See [ReactCommon\react-native-patched\ReactCommon\cxxreact\CxxNativeModule.cpp](https://github.com/facebook/react-native/blob/0b8a82a6eeeb3508b80ee137d313f64fe323db06/ReactCommon/cxxreact/CxxNativeModule.cpp#L97)). This breakpoint is bound to be hit a lot (every time a call to a native method is made), so we want to make sure we only break when *our* method of interest is involved.
```
NickGerleman, line 105: We probably don't want to point folks towards this. react-native-patched is a build time directory we're trying to remove (it will be gone on binary distribution as well), and Facebook has been refactoring some of this.

docs/native-code.md
asklar: The property exists, it is not on by default as it would bloat fully native apps unnecessarily.

docs/native-code.md
asklar: We can update the docs when the file gets removed.

stmoy (approved):

rectified95:
This a great addition - thanks doing it!

PR #271, opened 2020-09-23
Remove misleading bits from NM Docs

NickGerleman:
Remove the examples around "Methods that return more complex types (like std::string) can be implemented by returning void and accepting a bool and a ReactPromise<JSValue>:" which contradicts other docs and is a bit misinformed. Fixup some incorrect information about sync methods.


NickGerleman:
Closing since there is other work to revive these

PR #276, opened 2020-09-25
Add metro troubleshooting document

acoates-ms:
A common issue people hit is around their metro config.  This adds a starting set of steps for people to investigate their issue.

### DIFF
--- a/docs/metro-config-out-tree-platforms.md
+++ b/docs/metro-config-out-tree-platforms.md
@@ -1,0 +1,94 @@
+---
+id: metro-config-out-tree-platforms
+title: Metro config for out of tree platforms
+---
+
+How to setup and diagnose issues with metro config for `react-native-macos` and `react-native-windows`.  Many of the instructions within this document may refer just to `react-native-windows`, but the steps apply to `react-native-macos` in the same way.
+
+# Version requirements
+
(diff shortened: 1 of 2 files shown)


asklar:
We already have the metro troubleshooting guide that Keith and I have been maintaining, could we put this there instead, or better yet put everything in the metro project where it belongs?

acoates-ms:
Hmm.  I thought the RNW wiki was mostly for people developing within RNW.  Not for consumers of RNW.  This document is much more about trying to help people configure metro for their project, than how to debug or learn the internals of metro.

Also all of this applies to macOS too, and pointing macOS users to the RNW wiki seems like an odd place to point to.

I think maybe I can take some of the content from that wiki thats useful for consumers, then maybe we could change the aka.ms link to point to the webpage?

I was basically trying to write out the longer answer / guide for the constant issues we are getting on both RNW and RNMacOS repos where the basic metro config is clearly not quite setup for their specific repo setup.

chrisglein:
> Hmm. I thought the RNW wiki was mostly for people developing within RNW. Not for consumers of RNW. This document is much more about trying to help people configure metro for their project, than how to debug or learn the internals of metro.

Agreed that he wiki is targeted towards contributors where the website is targeted towards customers. Where we have a pain point for RNW customers we should get the right info on the website.

Issue #254, opened 2020-09-10
Document all annotations and APIs for Native Modules


stmoy:
@vmoroz indicated that not all annotations are listed in these tables:
- https://microsoft.github.io/react-native-windows/docs/native-modules#attributes
- https://microsoft.github.io/react-native-windows/docs/native-modules#attributes-1

We should make sure that the docs are up-to-date.

asklar:
@stmoy I think this bug should be about the annotations/macros and not the yarn versioning, please double check me :)

stmoy:
@asklar - correct, my bad. I updated my original issue with the correct links.

stmoy:
Would it be useful to separate the API docs vs. the guidance docs for attributes? Maybe each attribute could have its own page. 

Smallest thing would be to add the missing attributes REACT_INIT and REACT_FUNCTION.

asklar:
This should also speak about the differences and limitations between REACT_METHOD and REACT_SYNC_METHOD

asklar:
Needs documentation for C++ REACT_INIT

PR #280, opened 2020-10-01
update native module docs

### LINKED ISSUE #252: Update Native Module docs when returning complex types
The following feedback regards this [specific section](https://microsoft.github.io/react-native-windows/docs/native-modules#1-authoring-your-native-module-1) of the native modules docs, copied here for posterity:

-=-=-=-

Methods that return more complex types (like std::string) can be implemented by returning void and accepting a bool and a ReactPromise<JSValue>:

```
    REACT_METHOD(GetString, L"getString");
    void GetString(bool error, React::ReactPromise<React::JSValue>&& result) noexcept
    {
      if (error) {
        result.Reject("Failure");
      } else {
        std::string text = DoSomething();
        result.Resolve(React::JSValue{ text });
      }
    }
```

This can be also tied in with C++/WinRT event handlers or IAsyncOperation<T> like so:

```
    REACT_METHOD(GetString, L"getString");
    void GetString(bool error, React::ReactPromise<React::JSValue>&& result) noexcept
    {
      if (error) {
        result.Reject("Failure");
      } else {
        something.Completed([result] (const auto& status, const auto& operation) {
          // do error checking, e.g. status should be Completed
          std::wstring wtext{operation.GetResults()};
          result.Resolve(React::JSValue{ ConvertWideStringToUtf8String(text) });
        });
      }
    }
```

-=-=-=-

- This contradicts existing documentation and samples around ReactPromise and parameters
- We have built-in modules using returning Promise to string that don't accept extra booleans
- Following the instructions will cause a compile error when we enable spec checking
- What does "complex type" mean?

asklar:
Fixes #252


docs/native-modules.md
```diff
@@ -502,3 +513,7 @@ To access our `FancyMath` constants, we can simply call `NativeModules.FancyMath
...
 
 For events, you'll see that we created an instance of `NativeEventEmitter` passing in our `NativeModules.FancyMath` module, and called it `FancyMathEventEmitter`. We can then use the `FancyMathEventEmitter.addListener()` and `FancyMathEventEmitter.removeListener()` methods to subscribe to our `FancyMath::AddEvent`. In this case, when `AddEvent` is fired in the native code, `eventHandler` will get called, which logs the result to the console log.
+
+### JavaScript and Windows Runtime strings
+Note that JavaScript strings are UTF8 (i.e. `std::string`) but WinRT strings (`winrt::hstring` in C++/WinRT) are UTF16 (and therefore translate ot `std::wstring`), so when inter-operating between JavaScript and WinRT APIs, you will likely need to convert between these two formats.
```
stmoy, line 518: There's a lot of parens in this sentence and makes it a bit difficult to read. 

```suggestion
Note that JavaScript strings are UTF8 (i.e. `std::string`) but WinRT strings are UTF16 (ie. `winrt::hstring` in C++/WinRT which therefore translate to `std::wstring`), so when inter-operating between JavaScript and WinRT APIs, you will likely need to convert between these two formats.

```

stmoy (approved):
Overall looks good to me but I'd love somebody like @vmoroz to also sign off as well.

docs/native-modules.md
asklar: Good point :) I'll simplify it:

```suggestion
Note that JavaScript strings are UTF8 (i.e. `std::string`) but WinRT strings are UTF16 (i.e. `winrt::hstring` in C++/WinRT), so when inter-operating between JavaScript and WinRT APIs, you will need to convert between these two encodings.
```

website/versioned_docs/version-0.63/native-modules.md
```diff
@@ -542,3 +559,7 @@ To work around this problem there are three options:
       <DotNetNativeVersion>2.2.3</DotNetNativeVersion>
     </PropertyGroup>
     ```
+
+### JavaScript and Windows Runtime strings
+Note that JavaScript strings are UTF8 (i.e. `std::string`) but WinRT strings (`winrt::hstring` in C++/WinRT) are UTF16 (and therefore translate ot `std::wstring`), so when inter-operating between JavaScript and WinRT APIs, you will likely need to convert between these two formats.
```
asklar, line 564:
```suggestion
Note that JavaScript strings are UTF8 (i.e. `std::string`) but WinRT strings are UTF16 (i.e. `winrt::hstring` in C++/WinRT), so when inter-operating between JavaScript and WinRT APIs, you will need to convert between these two encodings.
```

docs/native-modules.md
```diff
@@ -293,17 +293,25 @@ Then we define constants, and it's as easy as creating a public field and giving
 
 It's just as easy to add custom methods, by attributing a public method with `REACT_METHOD`. In FancyMath we have one method, `add`, which takes two doubles and returns their sum. Again, we've specified the optional `name` argument in the `REACT_METHOD` macro-attribute so in JS we call `add` instead of `Add`.
 
-Methods that return more complex types (like `std::string`) can be implemented by returning void and accepting a bool and a `ReactPromise<JSValue>`:
+Native modules do not have the ability to check that the parameter types and the number of parameters match between what's called from JavaScript and what the native code accepts. However, the framework will validate that the number of promises-like parameters match: if the JavaScript API is async, it will expect that there is one "promise-like" parameter in the native method implementation signature.
```
vmoroz, line 296: >Native [](start = 0, length = 6)

I am not sure that it is correct to use the term 'promise-like'. It would be better to say that an asynchronous method has different way to return result. It can be one callback, two callbacks, or a promise.

docs/native-modules.md
```diff
@@ -293,17 +293,25 @@ Then we define constants, and it's as easy as creating a public field and giving
...
+See [Using Asynchronous Windows APIs](native-modules-async.md).
+
+Here is an example of an async method that returns a string. 
 
 ```cpp
     REACT_METHOD(GetString, L"getString");
-    void GetString(bool error, React::ReactPromise<React::JSValue>&& result) noexcept
+    void GetString(React::ReactPromise<std::string>&& result) noexcept
     {
       if (error) {
```
vmoroz, line 310: >error [](start = 10, length = 5)

Should we make the example more realistic? This variable is undefined. E.g. we can show this method calling some WinRT API.

docs/native-modules.md
```diff
@@ -312,21 +320,24 @@ This can be also tied in with C++/WinRT event handlers or `IAsyncOperation<T>` l
 
 ```cpp
     REACT_METHOD(GetString, L"getString");
-    void GetString(bool error, React::ReactPromise<React::JSValue>&& result) noexcept
+    void GetString(React::ReactPromise<std::string>&& result) noexcept
     {
       if (error) {
         result.Reject("Failure");
       } else {
         something.Completed([result] (const auto& operation, const auto& status) {
```
vmoroz, line 328: >something.Completed [](start = 8, length = 19)

Let's use the co-routine syntax instead.

docs/native-modules.md
```diff
@@ -312,21 +320,24 @@ This can be also tied in with C++/WinRT event handlers or `IAsyncOperation<T>` l
...
+          winrt::hstring result{operation.GetResults()};
+          result.Resolve(winrt::to_string(result));
         });
       }
     }
 ```
+See [JavaScript and Windows Runtime strings](#javascript-and-windows-runtime-strings) for more details.
 
-`REACT_SYNC_METHOD` isn't supported in Debug when the JS engine is Chrome V8. You will get an exception that reads:
+The [`JSValue`](native-modules-jsvalue.md) type can be used when the API returns a JavaScript objects or takes JavaScript objects as input parameters.
```
vmoroz, line 338: >The [](start = 0, length = 3)

I would rather say that the expectation is that the native module methods parameters are expected to be related to the problem domain. ReadValue and WriteValue function overrides can be implemented for custom types. In case if the strongly type approach is not possible, the developers can use the JSValue that is an immutable representation of a JSON-like objects.

docs/native-modules.md
```diff
@@ -312,21 +320,24 @@ This can be also tied in with C++/WinRT event handlers or `IAsyncOperation<T>` l
...
+
+Native modules will want to use `REACT_METHOD` instead of `REACT_SYNC_METHOD` since the latter precludes web debugging and has performance implications. When using web debugging you will see an exception that reads:
```
vmoroz, line 340: >performance implications [](start = 127, length = 24)

I do not think it is a true statement. On their own the synchronous methods do not cause perf issues. In future TurboModules all methods are planned to be synchronous according to the FB team. Should we rather speak about the differences and limitations instead of on insisting using the REACT_METHOD?

docs/native-modules.md
```diff
@@ -502,3 +513,7 @@ To access our `FancyMath` constants, we can simply call `NativeModules.FancyMath
 Calls to methods are a little different due to the asynchronous nature of the JS engine. If the native method returns nothing, we can simply call the method. However, in this case `FancyMath.add()` returns a value, so in addition to the two necessary parameters we also include a callback function which will be called with the result of `FancyMath.add()`. In the example above, we can see that the callback raises an Alert dialog with the result value.
 
 For events, you'll see that we created an instance of `NativeEventEmitter` passing in our `NativeModules.FancyMath` module, and called it `FancyMathEventEmitter`. We can then use the `FancyMathEventEmitter.addListener()` and `FancyMathEventEmitter.removeListener()` methods to subscribe to our `FancyMath::AddEvent`. In this case, when `AddEvent` is fired in the native code, `eventHandler` will get called, which logs the result to the console log.
+
+### JavaScript and Windows Runtime strings
+Note that JavaScript strings are UTF8 (i.e. `std::string`) but WinRT strings are UTF16 (i.e. `winrt::hstring` in C++/WinRT), so when inter-operating between JavaScript and WinRT APIs, you will need to convert between these two encodings.
```
vmoroz, line 518: >winrt::hstring [](start = 94, length = 14)

Since native module methods can support any parameter type which has ReadValue and WriteValue functions. we must fix this issue by adding the reader writer for winrt::hstring to avoid extra memory allocation due to the conversion.

vmoroz (approved):
:shipit:

docs/native-modules.md
asklar: Filed 6166

docs/native-modules.md
asklar: > I do not think it is a true statement. On their own the synchronous methods do not cause perf issues. In future TurboModules all methods are planned to be synchronous according to the FB team

I integrated this from @NickGerleman 's  [PR](https://github.com/microsoft/react-native-windows-samples/pull/271/files#diff-3faf77f987916cbf9f9a99f42e49b9d1R295) but I can take it out if it is wrong.

> Should we rather speak about the differences and limitations instead of on insisting using the REACT_METHOD?

Absolutely, we should explain the differences. I believe you have an issue filed on this: #254
Should I remove this paragraph and look to you to update this soon @vmoroz ?

docs/native-modules.md
asklar: I would love to point to the ReadValue and WriteValue documentation but we don't mention them anywhere in the docs. I've filed https://github.com/microsoft/react-native-windows-samples/issues/282 to add these APIs to the docs.

docs/native-modules.md
asklar: filed 283 to add an example using coroutines

docs/native-modules.md
asklar: @vmoroz We should:
- Define what a callback is. Is it strictly a `void (*) ()`? a `std::function`?  is it anything with an `operator()` ?
- What types they take/return
- What the difference between 1 or 2 is
- When you would use a callback, two callbacks or a promise

asklar: Filed 283

docs/native-modules.md
asklar: Add more realistic examples of native modules calling WinRT APIs
Including pointing to our sample native modules and possibly community modules

asklar: Filed 288

docs/native-modules.md
asklar: Filed 289

PR #295, opened 2020-10-06
Updated rssreader sample to React Native 0.62

qmatteoq:


jonthysell (approved):

jonthysell:
@qmatteoq Are you ready for this to be checked in? Would you like help giving it a CI loop?

qmatteoq:
@jonthysell that's a great idea! Do you have any guidance to follow for implementing a CI loop? I know how to do it with GitHub Actions, but I want to follow the approach you're already using for this project. Shall I go and build something similar to the workflows I see for the other samples?

jonthysell:
In this case, I would just copy the three yml files under [.github/workflows](https://github.com/microsoft/react-native-windows-samples/tree/master/.github/workflows) for say Calculator, rename with rssreader, and find/replace `Calculator` with `rssreader` in each file.

PR makes sure PRs don't break the sample, CI makes sure all flavors build after merges (and once a week just in case upstream dependencies break) and CI Upgrade simulates updating to newer versions of RNW once a week as a smoke test.

FYI: there are also some "generic" CI loops samples here: https://github.com/microsoft/react-native-windows-samples/tree/master/samples/ContinuousIntegration

They are sufficient for "regular" app repos out in the wild.

jonthysell:
I'm going to approve this, and we can add a CI loop in a new PR.

PR #302, opened 2020-10-21
Fix FancyMath.add to return via ReactPromise

ryfow:
Without this change, the method never actually gets called.


asklar:
this sounds similar to the issue I pointed out to @vmoroz when I was writing a native module. I wonder if that's a result of a bug on the NM infra side? CC @jonthysell too

ryfow:
Hmm, looks like maybe you're supposed to mark sync methods with `[ReactSyncMethod]`.  It makes sense that a sync method wasn't getting called when it's marked async.

ryfow:
I'm not blocked. I don't think the docs are right, but haven't tested to see if something changed in rn-windows to make it start working.

PR #312, opened 2020-11-09
Add docs for Hermes Consumption in RNW 0.64

NickGerleman:
The previous instructions didn't work outside of the react-native-windows repo, so new logic was added in 0.64 to allow properly opting into Hermes. These docs explain how to use it.

Validated everything looks correct when viewing the live preview of the docs site.

### DIFF
--- a/docs/hermes.md
+++ b/docs/hermes.md
@@ -65,6 +65,7 @@ Finally, edit the `ExtensionTargets` ItemGroup in your project's `vcxproj` file 
 If you've built a project using Hermes and want to opt-out, the above instructions may be reversed.
 
 ### Known limitations
+- Hermes does not yet support C# projects
 - Hermes debugger is not yet supported
 - Enabling Hermes leads to a very large download on initial build. This will become smaller in the future
 - Hermes is not yet compatible with other experimental features, like NuGet distribution or WinUI3
--- a/docs/hermes.md
+++ b/docs/hermes.md
@@ -65,7 +65,7 @@ Finally, edit the `ExtensionTargets` ItemGroup in your project's `vcxproj` file 
 If you've built a project using Hermes and want to opt-out, the above instructions may be reversed.
 
 ### Known limitations
-- Hermes does not yet support C# projects
+- Hermes is not yet supported when using C# projects
 - Hermes debugger is not yet supported
 - Enabling Hermes leads to a very large download on initial build. This will become smaller in the future
 - Hermes is not yet compatible with other experimental features, like NuGet distribution or WinUI3
--- a/docs/hermes.md
+++ b/docs/hermes.md
@@ -68,5 +68,5 @@ If you've built a project using Hermes and want to opt-out, the above instructio
 - Hermes is not yet supported when using C# projects
 - Hermes debugger is not yet supported
 - Enabling Hermes leads to a very large download on initial build. This will become smaller in the future
-- Hermes is not yet compatible with other experimental features, like NuGet distribution or WinUI3
+- Hermes is not yet compatible with other experimental features, like NuGet distribution or WinUI 3
 - The binary for Hermes is not yet signed by Microsoft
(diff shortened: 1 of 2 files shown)


asklar (approved):

docs/hermes.md
```diff
@@ -26,13 +26,48 @@ pod 'libevent', :podspec => '../node_modules/react-native-macos/third-party-pods
 > Be sure to set your target's deployment target to at least 10.14 before running `pod install`
 
 ## Hermes on Windows:
-We are working on bringing Hermes support to React Native for Windows through the official [Microsoft fork of Hermes](https://github.com/microsoft/hermes-windows).
+Hermes is experimentally supported on Windows, generally providing better performance characteristics than the default Chakra engine.
 
-Hermes for React Native for Windows has been updated to support the latest version 0.4.3 and can be turned on with a build flag for experimental use while we complete the feature set and get it ready for shipping.
+### Enabling Hermes for new projects
+The easiest way to enable Hermes is to pass the `--useHermes` flag to `react-native-windows-init` when creating a new project. This will set up your project to use the Herems engine and to generate bundles as Hermes bytecode instead of JavaScript.
```
asklar, line 32: The easiest way to enable Hermes is to pass the `--useHermes` flag to `react-native-windows-init` when creating a new project. This will set up your project to use the Hermes engine and to generate bundles as Hermes bytecode instead of JavaScript.

docs/hermes.md
```diff
@@ -26,13 +26,47 @@ pod 'libevent', :podspec => '../node_modules/react-native-macos/third-party-pods
...
+```
+
+### Disabling Hermes
+If you've built a project using Hermes and want to opt-out, the above instructions may be reversed.
+
+### Known limitations
+- Hermes is not yet supported when using C# projects
+- Hermes debugger is not yet supported
+- Enabling Hermes leads to a very large download on initial build. This will become smaller in the future
+- Hermes is not yet compatible with other experimental features, like NuGet distribution or WinUI3
```
asklar, line 71: Align with the rest of the docs:
```suggestion
- Hermes is not yet compatible with other experimental features, like NuGet distribution or WinUI 3
```

(and remove the Winui3 exemption)

docs/hermes.md
```diff
@@ -26,13 +26,47 @@ pod 'libevent', :podspec => '../node_modules/react-native-macos/third-party-pods
...
+  ...
+  <Import Project="..\packages\ReactNative.Hermes.Windows.0.7.1\build\native\ReactNative.Hermes.Windows.targets" Condition="Exists('..\packages\ReactNative.Hermes.Windows.0.7.1\build\native\ReactNative.Hermes.Windows.targets')" />
+</ImportGroup>
+```
+
+### Disabling Hermes
+If you've built a project using Hermes and want to opt-out, the above instructions may be reversed.
+
+### Known limitations
+- Hermes is not yet supported when using C# projects
```
asklar, line 68: Was the "make Hermes work in UWP" work completed? I see the Hermes task still marked untriaged: https://github.com/microsoft/hermes-windows/issues/20

Since UWP is the only supported target for RNW, I think this doc should be internal-only until we get UWP, otherwise it's going to be confusing for people building RNW [UWP] apps.

docs/hermes.md
```diff
@@ -26,13 +26,47 @@ pod 'libevent', :podspec => '../node_modules/react-native-macos/third-party-pods
...
+- Hermes debugger is not yet supported
+- Enabling Hermes leads to a very large download on initial build. This will become smaller in the future
+- Hermes is not yet compatible with other experimental features, like NuGet distribution or WinUI 3
+- The binary for Hermes is not yet signed by Microsoft
```
asklar, line 72: + Direct debugging is not supported, which means Turbo Modules are not supported either

docs/hermes.md
NickGerleman: My understanding is that @mganandraj stubbed out non-UWP APIs about a year ago, so we do compile with the correct app partition and everything works. I'm not sure we run WACK yet on Hermes, which we probably should.

But yeah, Hermes is focused solely UWP. Office/Win32 is invested in V8 in the near-term.

docs/hermes.md
NickGerleman: TurboModules can't yet be used outside of  RNW project internals in general, so this shouldn't be a concern.

PR #323, opened 2020-11-17
Mention VS 16.5 Requirements

NickGerleman:
https://github.com/microsoft/react-native-windows/pull/6548 means we will start requiring VS 16.5 or later. Add that as information to the docs.


asklar:
thanks @NickGerleman - mind adding this to 0.63 too since we're still supporting it?

asklar (approved):

NickGerleman:
@asklar I don't plan to backport the change to 0.63-stable, so this should only apply once we've released 0.64.

asklar (approved):

PR #325, opened 2020-11-19
Remove some guidance referencing internals that have changed

NickGerleman:
The section references a build-time directory that is no longer present, and some upstream implementation details that have been shifting quite frequently. Remove it.

### DIFF
--- a/docs/native-code.md
+++ b/docs/native-code.md
@@ -100,13 +100,6 @@ To work around this problem there are three options:
 
 So you added a new native module or a new method to a module but it isn't working, **now what?!**
 
-If your method isn't being hit in the VS debugger, something is blocking the call due to a mismatch, likely between the expected and actual types that your method takes/returns.
-
-To debug into what is rejecting the call, set a breakpoint in `CxxNativeModule::invoke` (See [`ReactCommon\react-native-patched\ReactCommon\cxxreact\CxxNativeModule.cpp`](https://github.com/facebook/react-native/blob/0b8a82a6eeeb3508b80ee137d313f64fe323db06/ReactCommon/cxxreact/CxxNativeModule.cpp#L97)). This breakpoint is bound to be hit a lot (every time a call to a native method is made), so we want to make sure we only break when *our* method of interest is involved.
-
-Right-click on the breakpoint to add a Condition. Suppose the method you are interested in catching is called `getString`. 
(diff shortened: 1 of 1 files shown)


asklar:
the file is still there, why are you proposing to remove this?

asklar (changes requested):
the file is still there on the core repo, seems like we should just fix the display name for the link instead of removing useful guidance

NickGerleman:
1. ReactCommon sources aren't user accessible on binary distribution
1. ReactCommon sources are already user hidden in a separate NPM package
1. When using JSI in the NM we won't go through this path
1. When TurboModules are enabled, we won't go through this path

This is going to change frequently and nobody is going to keep this up to date.

asklar:
> 1. ReactCommon sources aren't user accessible on binary distribution
> 2. ReactCommon sources are already user hidden in a separate NPM package
> 3. When TurboModules are enabled, we won't go through this path
> 
> This is going to change frequently and nobody is going to keep this up to date.

1. Nobody outside Office is using binary distribution today.
2. We can fix the text displayed in the link to point to where it would live in node_modules
3. We don't have any turbomodules for community modules, so this is not a problem today. We will have to update the docs anyway when this becomes an option.

NickGerleman:
The path to react-native in node_modules isn't stable either, but I think this is missing the meta-point of not wanting to document internals that change frequently. Especially ones that will be difficult to extract use out of. Office doesn't have NM2 yet at all, so this is strictly applicable to XAML. I can definitely understand wanting to help unblock users, but these docs already became unusable in a single release, and imo weren't actionable before. I only caught this because I remember asking we didn't document the react patching in the first place 🙂.

I know we've had some fundamentally different opinions on what these docs should look like, which is why I want to defer to the folks owning NM infrastructure. @vmoroz @acoates-ms do either of you have opinions on whether we should keep these documented?

asklar:
The content here is aimed at people using RNW (with XAML - which is what we call RNW). 
Usage specific to Office or other customers that aren't targeting XAML are out of scope.

NickGerleman:
> The content here is aimed at people using RNW (with XAML - which is what we call RNW).
> Usage specific to Office or other customers that aren't targeting XAML are out of scope.

Reiterating, this doesn't apply to Office at all, since we aren't able to use native modules 2.0 yet.

NickGerleman:
stale pr

PR #329, opened 2020-11-23
Fix async native module docs

jonthysell:
Addresses issues in https://github.com/microsoft/react-native-windows/issues/6563 by restoring docs to match the working samples.

### DIFF
--- a/docs/native-modules-async.md
+++ b/docs/native-modules-async.md
@@ -25,94 +25,122 @@ NativeModules.SimpleHttpModule.GetHttpResponse('https://microsoft.github.io/reac
 
 ### `SimpleHttpModule` in C#
 
-The native module support for C# supports the common asynchronous programming patterns established in C# using `async`, `await` and `Task<T>`.
+Let's start with the asynchronous native method which performs the HTTP request:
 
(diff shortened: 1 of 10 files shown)


samples/NativeModuleSample/csharp/package.json
```diff
@@ -30,6 +30,6 @@
   "devDependencies": {
     "react": "^16.13.1",
     "react-native": "^0.63.2",
-    "react-native-windows": "^0.63.0-0"
+    "react-native-windows": "0.63.10"
```
asklar, line 33: shouldn't this be ^0.63.10 so we take future updates?

docs/native-modules-async.md
```diff
@@ -25,94 +25,122 @@ NativeModules.SimpleHttpModule.GetHttpResponse('https://microsoft.github.io/reac
...
-      
-      return content;
+  // Build result object
+  var resultObject = new JSValueObject();
+
+  resultObject["statusCode"] = (int)statusCode;
+  resultObject["content"] = content;
+
+  promise.Resolve(resultObject);
+}
```
NickGerleman, line 105: This isn't right... This has a return type of Task but doesn't return anything. @dannyvv did some work to support task-style returns before. Let's make sure to loop him in.

PR #351, opened 2020-12-22
Update Versions in CI Loop Sample Code

chiaramooney:
Updated versions for Nuget and Msbuild in .yml code snippet. These were the versions I needed to update to for CI Loop to work on RN Gallery. Old versions apparently use deprecated calls.

### DIFF
--- a/docs/native-modules-setup.md
+++ b/docs/native-modules-setup.md
@@ -176,12 +176,12 @@ Next you need to create a YAML file for GitHub Actions, the basic steps are:
         node-version: '12.9.1'
 
     - name: Setup MSBuild
-      uses: microsoft/setup-msbuild@v1.0.0
+      uses: microsoft/setup-msbuild@v1
       with:
         vs-version: 16.5
 
     - name: Setup NuGet
-      uses: NuGet/setup-nuget@v1.0.2
+      uses: NuGet/setup-nuget@v1
 
     - name: Check node modules cache
       uses: actions/cache@v1
--- a/website/versioned_docs/version-0.63/native-modules-setup.md
+++ b/website/versioned_docs/version-0.63/native-modules-setup.md
@@ -399,12 +399,12 @@ Next you need to create a YAML file for GitHub Actions, the basic steps are:
         node-version: '12.9.1'
 
     - name: Setup MSBuild
-      uses: microsoft/setup-msbuild@v1.0.0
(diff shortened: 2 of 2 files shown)


kmelmon (approved):
:shipit:

docs/native-modules-setup.md
```diff
@@ -176,12 +176,12 @@ Next you need to create a YAML file for GitHub Actions, the basic steps are:
...
 
     - name: Setup MSBuild
-      uses: microsoft/setup-msbuild@v1.0.0
+      uses: microsoft/setup-msbuild@v1
       with:
         vs-version: 16.5
 
     - name: Setup NuGet
-      uses: NuGet/setup-nuget@v1.0.2
+      uses: NuGet/setup-nuget@v1
```
asklar, line 184: If this is targeting 0.64+, you should be able to delete this now, we use msbuild's built-in nuget support now

website/versioned_docs/version-0.63/native-modules-setup.md
```diff
@@ -399,12 +399,12 @@ Next you need to create a YAML file for GitHub Actions, the basic steps are:
...
 
     - name: Setup MSBuild
-      uses: microsoft/setup-msbuild@v1.0.0
+      uses: microsoft/setup-msbuild@v1
       with:
         vs-version: 16.5
        
     - name: Setup NuGet
-      uses: NuGet/setup-nuget@v1.0.2
+      uses: NuGet/setup-nuget@v1
```
asklar, line 407: do we need to update 0.64 and next?

docs/native-modules-setup.md
chiaramooney: Ah so we can remove the setup Nuget all together for 64+ then?

website/versioned_docs/version-0.63/native-modules-setup.md
chiaramooney: It doesn't look like we have versioned docs for 0.64 yet. If the above is true and we are using msbuild's built-in nuget support then the change I made in next should be removed.

docs/native-modules-setup.md
asklar: yes, see #6548

chiaramooney: Changes made in PR #352

PR #359, opened 2021-01-26
Revert "Fix async native module docs"

### LINKED ISSUE #340: Re-add docs  and add samples for task based C# native modules
Offline discussion was that https://github.com/microsoft/react-native-windows-samples/pull/329 was the wrong change. We updated C# native modules to support Tasks, but samples still used ReactPromise. It was thought that docs were out of date, so docs were updated to refer to the old method used by samples instead of the other way around.

NickGerleman:
Reverts microsoft/react-native-windows-samples#329

Fixes #340

Task based promises are the preferred solution to ReactPromise. Docs were accidentally updated to refer to the old way, which was being used by samples. Restore docs for Task based async NMs.

### DIFF
--- a/docs/native-modules-async.md
+++ b/docs/native-modules-async.md
@@ -25,122 +25,94 @@ NativeModules.SimpleHttpModule.GetHttpResponse('https://microsoft.github.io/reac
 
 ### `SimpleHttpModule` in C#
 
-Let's start with the asynchronous native method which performs the HTTP request:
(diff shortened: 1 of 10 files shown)


NickGerleman:
I missed that this commit also changed package versions, which we don't want to revert.

dannyvv:
@NickGerleman Was about to say to let yours go through first, I didn't realize I should update 0.63 too...
I'll update my pr...

PR #368, opened 2021-01-29
Evergreen version numbers for NM Setup Guide

NickGerleman:
The community module setup guide in versioned docs references specific version numbers, but instead calls for canary in the unversioned docs. Replace it with something generic that could be used across new versions if we don't update it.


dannyvv (approved):
:shipit:

asklar (approved):

PR #370, opened 2021-01-29
Update Hermes NuGet Package Version

NickGerleman:
We don't really have channels or anything yet. Keeping hard version numbers until Hermes NuGet package is downloaded by default and we don't need to worry about this.


dannyvv (approved):
:shipit:

PR #373, opened 2021-01-29
Remove SetValue usage

asklar:
Fixes non-static usage of `DependencyObject.SetValue`


chrisglein:
> Fixes non-static usage of `DependencyObject.SetValue`

Happen to know why it was this way before?
There's no linked issue here so can you flesh out a bit more of the "why" in this PR description?
Looking at the change itself... looks just clearly better :'D

chrisglein (approved):

PR #384, opened 2021-02-10
Update docs to point to react-native-macos 0.63 as latest

acoates-ms:


acoates-ms:
Unfortunately the link this is pointing to wont exist until we release the 0.64 docs.  I think maybe I want to wait on this until we bump the docs.

chrisglein (approved):

asklar:
@acoates-ms Does this close #399  ?

PR #386, opened 2021-02-11
Update Hermes Instructions

NickGerleman:
Depends on https://github.com/microsoft/react-native-windows/pull/7101/files

This updates Hermes instructions for new build logic, changes headings between macOS and Windows to be consistent, and moves the macOS section below Windows, since Windows has more usage.

### DIFF
--- a/docs/hermes.md
+++ b/docs/hermes.md
@@ -8,23 +8,6 @@ The [Hermes](https://hermesengine.dev/) engine is an open source JavaScript engi
 
 To learn more about what it is and how to use it, check out the [React Native](https://reactnative.dev/docs/hermes#docsNav) documentation for it.
 
-## Available on macOS
-Hermes is available on React Native for macOS, provided you are using version 0.62 or higher.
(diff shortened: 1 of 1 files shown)


docs/hermes.md
```diff
@@ -34,10 +17,10 @@ The easiest way to enable Hermes is to pass the `--useHermes` flag to `react-nat
 ### Using Hermes in an existing project
 Hermes can be enabled in existing projects with a few edits.
 
-First, set the `UseHermes` property to `true` in the `BuildFlags.props` file in your project's `windows` directory:
+First, set the `UseHermes` property to `true` in the `ExperimentalFeatures.props` file in your project's `windows` directory:
```
asklar, line 20: maybe it would be good to make useHermes behave the same way as useWinUI3: you can set it in react-native.config and it will update the props file

NickGerleman: My hope is we can get the NuGet package slim enough in the future to include by default. That way there would only be a single switch needed

NickGerleman:
PR renaming props merged into 0.64-stable. This should be ready to merge.

asklar (approved):

PR #383, opened 2021-02-10
v 0.64 modules update and TOC cleanup

stmoy:
Updating the list of modules with Windows implementations and cleaning up the table of contents.

PR #390, opened 2021-02-17
Update supported-community-modules.md

asklar:
fixing spellchecker errors


asklar:
@stmoy  fyi this came from #383, the CI was complaining because rn in rn-fetch-blob and tts and pdf are not words, so we just mark those as code to avoid the spellchecker

stmoy (approved):

stmoy:
Got it, thanks for the FYI

AgneLukoseviciute (approved):

PR #395, opened 2021-02-18
Update Website to Include Wrapped Windows Control Blurb

chiaramooney:
Update website to include blurb which discusses what wrapped Windows Controls are and when to build UI components with this strategy vs. using JS. 

Currently a draft, open to edits and additions. Will updated old versioned docs after blub is finalized.


docs/view-managers.md
```diff
@@ -28,6 +28,11 @@ Currently, React Native for Windows uses a global `PerspectiveTransform` to prov
 
 __Important__: The `IsPerspectiveEnabled` property is experimental and support for it may be removed in the future.
 
+### Note about wrapping Windows controls
+As described above, some native UI components such as `ScrollView` and `TextInput` have been built by wrapping native Windows controls from the UWP XAML controls library. As a result, their visual appearance, animations, etc. will always match their native Windows counterpart. UI components can also be built using Javascript. In this case, the component's source code does not wrap a native control, but rather combines already contructed React Native Windows components to build itself. Components built in this way contain no C++/C# code.
```
stmoy, line 32:
```suggestion
As described above, some native UI components such as `ScrollView` and `TextInput` have been built by wrapping native Windows controls from the UWP XAML controls library. As a result, their visual appearance, animations, user experience, accessibility, etc. will always match their native Windows counterpart. 

UI components can also be built using Javascript. In this case, the component's source code does not wrap a native control, but rather combines already constructed React Native Windows components to build itself. Components built in this way contain no native code.

```

stmoy: It'll probably be a good idea to have some blurb about the various pros/cons of the two approaches. For example, wrapping will provide full-fidelity native code, but trade-off cross-platform user experience. JS controls can be service-delivered and styled using CSS concepts, but lack full native fidelity and require you to build your own accessibility solutions.

asklar:
```suggestion
As described above, some native UI components such as `ScrollView` and `TextInput` have been built by wrapping native Windows controls from the UWP XAML controls library. As a result, their visual appearance, animations, user experience, accessibility, etc. will always match their native Windows counterpart. 

UI components can also be built using JavaScript. In this case, the component's source code does not wrap a native control, but rather combines already constructed React Native Windows components to build itself. Components built in this way contain no native code. However, components built this way may not be as fully-featured or match the look and feel of the operating system the app is running on.
```

docs/view-managers.md
```diff
@@ -28,6 +28,13 @@ Currently, React Native for Windows uses a global `PerspectiveTransform` to prov
 
 __Important__: The `IsPerspectiveEnabled` property is experimental and support for it may be removed in the future.
 
+### Note about wrapping Windows controls
+As described above, some native UI components such as `ScrollView` and `TextInput` have been built by wrapping native Windows controls from the UWP XAML controls library. As a result, their visual appearance, animations, user experience, accessibility, etc. will always match their native Windows counterpart. 
+
+UI components can also be built using JavaScript. In this case, the component's source code does not wrap a native control, but rather combines already constructed React Native Windows components to build itself. Components built in this way contain no native code.
```
asklar, line 34: does not _necessarily_/_directly_ wrap a native control
something will eventually draw some pixels on screen, through a native ui control, so at some point someone wraps a native control (unless people are directly drawing with win2d)

docs/view-managers.md
```diff
@@ -28,6 +28,13 @@ Currently, React Native for Windows uses a global `PerspectiveTransform` to prov
...
 
+### Note about wrapping Windows controls
+As described above, some native UI components such as `ScrollView` and `TextInput` have been built by wrapping native Windows controls from the UWP XAML controls library. As a result, their visual appearance, animations, user experience, accessibility, etc. will always match their native Windows counterpart. 
+
+UI components can also be built using JavaScript. In this case, the component's source code does not wrap a native control, but rather combines already constructed React Native Windows components to build itself. Components built in this way contain no native code.
+
+If you are looking for your constructed component to always match the styling of its counterpart in Windows, choosing to build your component by wrapping a XAML control would be the best option for you. Follow the steps below to learn how to build your native UI component. Conversely, if this characteristic is not important for your component's purpose, construct the component in JavaScript using other React Native Windows components as building blocks. Follow the steps [here](https://reactnative.dev/docs/intro-react) to learn how to build your component.
```
asklar, line 36: I wonder if we should clarify that if you want to use vanilla XAML controls without any extensive customization you can use the react-native-xaml module

docs/view-managers.md
chiaramooney: Ok! I took the paragraph on line 32 from the tooltip statement we landed on for the "Wrapped Windows Control" badge in gallery. Should I be scrapping that paragraph and replacing with ^^? Or leave it and then follow with the note you mentioned above?

acoates-ms: TextInput at least shouldn't match the windows visual appearance.  TextInput in react-native is a primitive concept, not a full UI component.  -- And as such it doesn't match the windows visual appearance, as it shouldn't have any border etc like the native control has.

asklar: Yes, this is an important point.

@chiaramooney / @stmoy  maybe it would be worth adding a page with a table of which UI components in RNW just wrap the native control versus not?

asklar:
@chiaramooney looks like this got closed due to inactivity. Was there something you needed to unblock?

chiaramooney:
> @chiaramooney looks like this got closed due to inactivity. Was there something you needed to unblock?
Ah oops yes I've put this PR on hold because it seems that there's more to the story of what a "wrapped windows control" means than we initially thought. I'll go ahead and close this PR for now and come back once I've finished by controls audit.

PR #398, opened 2021-03-03
Fix missing VS dep info

rectified95:
In the list of manual installation steps, we don't mention a VS component that needs to be installed in order to build the project.

### DIFF
--- a/docs/rnw-dependencies.md
+++ b/docs/rnw-dependencies.md
@@ -30,6 +30,7 @@ Alternatively, you can setup your environment manually:
 - Install [Visual Studio 2019 (version 16.5 or greater)](https://www.visualstudio.com/downloads) **with the following options checked**:
   - Workloads
     - Universal Windows Platform development
+      - Include C++ (v142) Universal Windows Platform tools (under 'Optional')
     - Desktop development with C++
     - .NET Desktop development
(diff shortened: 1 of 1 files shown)


asklar:
do we need to update the rnw-dependencies.ps1 script or is it already ok?

docs/rnw-dependencies.md
```diff
@@ -30,6 +30,7 @@ Alternatively, you can setup your environment manually:
 - Install [Visual Studio 2019 (version 16.5 or greater)](https://www.visualstudio.com/downloads) **with the following options checked**:
   - Workloads
     - Universal Windows Platform development
+      - Include C++ (v142) Universal Windows Platform tools (under 'Optional')
```
asklar, line 33: you can silence the spellchecker by quoting this:
```suggestion
      - Include `C++ (v142) Universal Windows Platform tools` (under 'Optional')
```

asklar (approved):

rectified95:
The script contains this package, and mentions it's optional for MSBuild, but required for VS - 'Microsoft.VisualStudio.ComponentGroup.UWP.VC'
Doesn't contain the v142 string, though
I used the script some time ago and it worked, so I think it's fine.

PR #403, opened 2021-03-11
Add 0.63 WinRT API docs

asklar:
These won't have any hand-written info since we only started using DOC_STRING in 0.64
However they are still valuable as they document what types exist, APIs, etc.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/403)


website/.unbroken_exclusions
```diff
@@ -1,6 +1,7 @@
 !node_modules
 !blog/2019-*
 !blog/2020-*
+!versioned_docs/version-0.63/*-api-windows*.md
```
jonthysell, line 4: Can you update the fix-unbroken.js script with these exclusions?

https://github.com/microsoft/react-native-windows-samples/blob/66e09b9da02690c360ec40716091463a6ced7c31/website/fix-unbroken.js#L70

jonthysell: Otherwise the next time we run the script, it will get lost.

asklar: yeah; maybe we just always insert `!versioned_docs/*/*-api-windows*.md` ?

jonthysell (approved):

PR #405, opened 2021-03-11
Update fix-unbroken.js to load existing file, add PR to check file is up to date

jonthysell:
Docusarus rewrites the relative URLs for markdown files, but not for assets. This causes unbroken to freak out. So we have the fix-unbroken.js script to generate an exclusions file so unbroken only catches "real" unbroken links.

The script should be run any time someone changes links, add/removed markdown files or assets, etc, however people constantly forget, and their PR fails. Or users add exclusions manually to the file, and then the script overwrites them the next time it's run.

This PR updates fix-unbroken.js to read in the list of statically define exclusions from the existing file and keep them in the newly generated file. It also adds a check to the Website PR to make sure that the file is up to date, and more specifically, give a useful error message so people re-run the script.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/405)

### DIFF
--- a/website/fix-unbroken.js
+++ b/website/fix-unbroken.js
@@ -66,8 +66,8 @@ var exclusions = [];
 // Load existing base exclusions
 const existingExclusions = fs.readFileSync('.unbroken_exclusions').toString().split(/\r?\n/) || [];
 
-for (var i = 0; i < existingExclusions.length; i++) {
-    const exclusion = existingExclusions[i].trim();
+for (let exclusion of existingExclusions) {
+    exclusion = exclusion.trim();
 
     if (exclusion === '') {
         exclusions.push(exclusion);
--- a/website/fix-unbroken.js
+++ b/website/fix-unbroken.js
@@ -63,11 +63,24 @@ console.log('Generating exclusions...');
 
 var exclusions = [];
 
-// Base exclusions
-exclusions.push(normalizePath('!node_modules'));
-exclusions.push(normalizePath('!blog\\2019-*'));
-exclusions.push(normalizePath('!blog\\2020-*'));
-exclusions.push('');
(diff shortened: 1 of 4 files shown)


website/fix-unbroken.js
```diff
@@ -63,11 +63,24 @@ console.log('Generating exclusions...');
...
 
-// Base exclusions
-exclusions.push(normalizePath('!node_modules'));
-exclusions.push(normalizePath('!blog\\2019-*'));
-exclusions.push(normalizePath('!blog\\2020-*'));
-exclusions.push('');
+// Load existing base exclusions
+const existingExclusions = fs.readFileSync('.unbroken_exclusions').toString().split(/\r?\n/) || [];
+
+for (var i = 0; i < existingExclusions.length; i++) {
```
asklar, line 69: nit, `let`, or even better `for...of`

website/fix-unbroken.js
```diff
@@ -63,11 +63,24 @@ console.log('Generating exclusions...');
...
+    const exclusion = existingExclusions[i].trim();
+
+    if (exclusion === '') {
+        exclusions.push(exclusion);
+    }
+    else if (exclusion.startsWith('#fix-unbroken.js')) {
+        exclusions.push(exclusion);
+        break;
```
asklar, line 77: i may be missing something, wouldn't this just remove everything below the `#` line next time this runs?

jonthysell: The items above the line are manual "static" exclusions, like the old blog posts. Everything below the line gets calculated by the script, which it does after the for loop.

PR #406, opened 2021-03-12
Update description of "--overwrite" flag

NickGerleman:
The referenced issue is no longer re levant, but we still use a windows specific blocklist. Remove the stale issue, and fix some typos.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/406)

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -36,7 +36,7 @@ Lastly, install the React Native for Windows packages.
 npx react-native-windows-init --overwrite
 ```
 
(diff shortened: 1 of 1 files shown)


asklar (approved):

asklar:
does this need to be ported to 64/63 docs?

NickGerleman:
0.64 docs haven't been cut yet, so this would go into them. Re 0.63 docs, this could go there, but is mostly just polish, so I don't think it had to.

PR #443, opened 2021-05-23
add note to include 1-2 Windows 10 SDK versions back

brodybits:
in docs/rnw-dependencies.md

as needed to avoid build errors due to inconsistent Windows 10 SDK version
between react-native-windows and recent Visual Studio 2019 version

related to issue: https://github.com/microsoft/react-native-windows/issues/4960

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/443)

PR #444, opened 2021-05-23
Overhaul Manual Setup info in rnw-dependencies.md

brodybits:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/444)

### DIFF
--- a/docs/rnw-dependencies.md
+++ b/docs/rnw-dependencies.md
@@ -32,6 +32,7 @@ Alternatively, you can setup your environment manually:
   - Workloads
     - Universal Windows Platform development
       - Include `C++ (v142) Universal Windows Platform tools` (under 'Optional')
+      - Older Windows 10 SDK version may be needed at this point.
     - Desktop development with C++
     - .NET Desktop development
   - Individual Components
--- a/docs/rnw-dependencies.md
+++ b/docs/rnw-dependencies.md
@@ -36,10 +36,7 @@ Alternatively, you can setup your environment manually:
   - Individual Components
     - Development activities
       - Node.js development support (optional; included by the `Node.js development` workload; not needed if Node.js is installed in a different way)
-
-You will also need to ensure that certain settings are enabled:
-- Long path support
-- Developer mode enabled
+- Ensure that long path support is enabled.
 
 #### React Native Development Dependencies
 
--- a/docs/rnw-dependencies.md
+++ b/docs/rnw-dependencies.md
@@ -54,6 +54,10 @@ Alternatively, you can setup your environment manually:
 
 - Install [Chrome](https://www.google.com/chrome/) (_optional_, but needed for JS debugging)
 - Install [Yarn](https://yarnpkg.com/en/docs/install) (_optional_ if only consuming react-native-windows, but **required** to contribute to react-native-windows)
+- *Highly recommended:* Install `git` using a method such as:
+  - Using a package manager such as [Chocolatey](https://chocolatey.org/) or [Scoop](https://scoop.sh/)
+  - Install [git for Windows](https://gitforwindows.org/)
+  - Install [GitHub Desktop](https://desktop.github.com/)
 
 ### Troubleshooting
 
--- a/docs/rnw-dependencies.md
+++ b/docs/rnw-dependencies.md
@@ -49,6 +49,7 @@ You will also need to ensure that certain settings are enabled:
   ```bat
   choco install nodejs-lts
   ```
+  - Using [another package manager](https://nodejs.org/en/download/package-manager/) such as [Scoop](https://scoop.sh/) or [Node Version Switcher (nvs)](https://github.com/jasongin/nvs)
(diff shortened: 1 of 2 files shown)


asklar (approved):

docs/rnw-dependencies.md
```diff
@@ -35,7 +35,7 @@ Alternatively, you can setup your environment manually:
     - .NET Desktop development
   - Individual Components
     - Development activities
-      - Node.js development support (optional)
+      - Node.js development support (optional; included by the `Node.js development` workload)
```
brodybits, line 38:
```suggestion
      - Node.js development support (optional; included by the `Node.js development` workload; not needed if Node.js is installed in a different way)
```

brodybits:
This proposal has turned into more of an overhaul, and I have done some testing with Windows Server 2019 on the cloud since my Windows 10 laptop is in pretty bad shape. I think and hope this can improve the situation for newbies in the future.

I think using [Scoop](https://scoop.sh/) can reduce if not eliminate the need for the elevated (Power)Shell during the manual setup.

docs/rnw-dependencies.md
```diff
@@ -27,33 +27,37 @@ Start an **elevated** PowerShell window and run:
 
 Alternatively, you can setup your environment manually:
 - Ensure Developer Mode is turned ON in Windows Settings App.
-- Install [Visual Studio 2019 (version 16.5 or greater)](https://www.visualstudio.com/downloads) **with the following options checked**:
-  - Workloads
+- It is _highly_ recommended to update the Windows system.
+- Install a recent version of [Visual Studio 2019](https://www.visualstudio.com/downloads) **with the following options checked**:
```
brodybits, line 31: This change is in response to a statement in <https://github.com/microsoft/react-native-windows/pull/7779#issuecomment-846612551>:

> [...] so we stop building with older compiler/linkers pretty soon after a VS release is marked as public. In reality we can only ever support the latest VS.

docs/rnw-dependencies.md
```diff
@@ -27,33 +27,37 @@ Start an **elevated** PowerShell window and run:
...
+  - **Workloads**
     - Universal Windows Platform development
       - Include `C++ (v142) Universal Windows Platform tools` (under 'Optional')
+      - Older Windows 10 SDK version may be needed at this point.
```
brodybits, line 35: This statement is a more fuzzy version of what I proposed in PR #443.

I think this statement is needed at least until the proposal in <https://github.com/microsoft/react-native-windows/pull/7779> is merged and released.

Even after <https://github.com/microsoft/react-native-windows/pull/7779> is released, I think this statement would be needed in case Visual Studio suddenly updates the default Windows 10 SDK version again.

brodybits (commented):
I am proposing a couple of fuzzy statements in response to the challenge of keeping up with the "aggressive" VS 2019 updates that affect the default Windows 10 SDK version.

NickGerleman (approved):
Looks like a great change! Thanks for the contribution.

docs/rnw-dependencies.md
NickGerleman: I agree with this change. There was some recent internal discussion where we discovered VS 16.6 no longer built our master branch correctly. We don't really test older versions, so our stated support level really does have to be evergreen.

jonthysell (approved):

docs/rnw-dependencies.md
brodybits: It looks like they are updating VS again & again: https://devblogs.microsoft.com/visualstudio/visual-studio-2019-v16-10-and-v16-11-preview-1-are-available-today/

docs/rnw-dependencies.md
```diff
@@ -27,33 +27,37 @@ Start an **elevated** PowerShell window and run:
...
-#### React Native Development Dependencies
-
-- Install the [standard React Native dependencies](https://reactnative.dev/docs/environment-setup)
-- Install [Node.js](https://nodejs.org) via one of the following methods:
+Options to install [Node.js](https://nodejs.org) separately:
   - Using [Chocolatey](https://chocolatey.org/) (_React Native recommended_). To use chocolatey, from an elevated Command Prompt, run:
   ```bat
   choco install nodejs-lts
   ```
+  - Using [another package manager](https://nodejs.org/en/download/package-manager/) such as [Scoop](https://scoop.sh/) or [Node Version Switcher (nvs)](https://github.com/jasongin/nvs)
```
brodybits, line 50: This line seems to cause the build failure, don't know how we should fix it.

AgneLukoseviciute: Markdown-spellcheck likely triggered by nvs -- if you can capitalize it, I think it will be treated as an acronym and be ignored.

brodybits:
I just pushed a couple more updates, don't know if it will need a re-review.

jonthysell (approved):

PR #467, opened 2021-07-05
Add copy button to all code blocks as recently documented by docusaurus

asklar:
This gets rid of the one off hacky copy button in rnw-dependencies and uses the recommended way to add the copy button to code blocks - see https://gist.github.com/yangshun/55db997ed0f8f4e6527571fc3bee4675
Also created an aka.ms redirect for the dependencies ps1 so it looks neater on the website, it's easier to remember, and allows us to move the file if we ever need to : )

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/467)

### DIFF
--- a/website/siteConfig.js
+++ b/website/siteConfig.js
@@ -66,7 +66,13 @@ const siteConfig = {
   scripts: [
     "https://cdn.jsdelivr.net/npm/focus-visible@5.0.2/dist/focus-visible.min.js",
     "https://platform.twitter.com/widgets.js",
-    "https://buttons.github.io/buttons.js"
+    "https://buttons.github.io/buttons.js",
+    'https://cdnjs.cloudflare.com/ajax/libs/clipboard.js/2.0.0/clipboard.min.js',
+    '/react-native-windows/js/code-block-buttons.js',
+  ],
+
+  stylesheets: [
+    '/react-native-windows/css/code-block-buttons.css',
   ],
 
   // On page navigation for the current documentation page.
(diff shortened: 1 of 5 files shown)


website/siteConfig.js
```diff
@@ -66,7 +66,13 @@ const siteConfig = {
   scripts: [
     "https://cdn.jsdelivr.net/npm/focus-visible@5.0.2/dist/focus-visible.min.js",
     "https://platform.twitter.com/widgets.js",
-    "https://buttons.github.io/buttons.js"
+    "https://buttons.github.io/buttons.js",
+    'https://cdnjs.cloudflare.com/ajax/libs/clipboard.js/2.0.0/clipboard.min.js',
+    '/react-native-windows/js/code-block-buttons.js',
+  ],
+
+  stylesheets: [
```
jonthysell, line 74: Does this not get included automatically just be adding it into the static css folder?

jonthysell (approved):

website/siteConfig.js
asklar: apparently not 🙃 the docusaurus docs point to that gist and the gist said to include it. @slorber do you know?

slorber: I don't really know 😅 started working on Docusaurus mostly on v2 and all this is quite old. I guess the gist it right

PR #478, opened 2021-07-15
Fix Highlighting of Misspelled Words in CI Output 

NickGerleman:
This repo uses markdown-spellcheck to catch document spelling errors. Its output shows the word it does not recognize via coloration. It uses Chalk to do this, which will by default noop when running in a non-interactive console (i.e. something that doesn't know how to show color).

This lack of color means we can't tell the word being complained about. GitHub Actions can render color, but we need to force chalk to. We pass the `--color` CLI arg for chalk to read, and override default logic to omit ANSI escape codes.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/478)

### DIFF
--- a/.github/workflows/website-pr.yml
+++ b/.github/workflows/website-pr.yml
@@ -19,7 +19,7 @@ jobs:
       uses: actions/checkout@v2
 
     - name: Spell-check docs
-      run: npx markdown-spellcheck "docs\*.md" "!docs\*-api-windows.md" --en-us --ignore-acronyms --ignore-numbers --report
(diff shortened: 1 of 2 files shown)


NickGerleman:
Confirmed fixed 
image

AgneLukoseviciute (approved):

asklar (approved):
nice, thanks!

dannyvv (approved):

PR #507, opened 2021-08-05
fix: template literal typo

akulsr0:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/507)


asklar:
I'm fairly certain the code is correct as it is, since json does not support interpolated strings. VSCode is responsible for re-interpreting the `${workspace}` string as an implicit interpolation.

PR #518, opened 2021-08-21
Showcase page with mock data

asklar:


chiaramooney (approved):
👍

stmoy:
> I get the need for a showcase page - but do we want to wait to expose this until we have actual apps to show? Or maybe add a "coming soon" at the top and use lorem ipsom text?

Agreed. I believe that Alex posted this to get feedback on the format/layout, but I am working on putting together the actual content. I'm not sure if Alex was intending to actually host the website with mock data, but I don't think there is much value in doing that.

asklar:
Pinging this so the bot doesn't age it out :) @stmoy lmk when you will have the actual data

PR #524, opened 2021-08-25
Lowercase bug docs update

HeyImChris:
There's a bug on macOS right now where you'll hit a registration error if you follow the steps in the docs. The issue is that it fails with casing differences and defaults to all lowercase.

We should fix this but in the meantime, let's update our docs so new developers can be easily unblocked.

I created this [GH issue](https://github.com/microsoft/react-native-macos/issues/819) to track it

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/524)


pkubat:
I still don't see this updated in the getting started guide.  If you follow the directions and use RN recommend naming you to create a project that does not load.  I just wasted hours to find this issue and a sure quick fix. As a RN developer I found this very frustrating

HeyImChris:
Digging into the docs here and it was added to the 0.63 Getting Started docs as that was the latest version when I filed the bug.

It _didn't_ get added to the 0.64 docs when those were created a few weeks ago. I'll look into updating those docs with the recommended workaround.

https://microsoft.github.io/react-native-windows/docs/0.63/rnm-getting-started

PR #536, opened 2021-09-13
Update NativeModuleSample to use 0.64 library templates and add upgrade instructions

jonthysell:
NativeModuleSample was created before the cli templates were added to RNW 0.64 and exposed through `npx react-native-windows-init --projectType lib`.

This PR updates the samples to align with the template, to facilitate upgrades to newer versions of RNW using the newer templates. It also adds setup, build, and upgrade steps to the readmes.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/536)


jonthysell:
This is in preparation of upgrading the sample to 0.65 (and documenting the process for the next upgrade).

PR #539, opened 2021-09-17
Update hermes.md

taenri:
Add an additional Known Limitation to point out that Hermes cannot be used at the same time as the Web Debugger.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/539)


dannyvv (approved):

NickGerleman:
I'm not sure this makes sense to call out. It's not a unique limitation to Hermes, at least.

asklar:
agree that this clarification isn't specific to hermes. A better place to put this doc is in the API doc for JSEngine and related.

asklar (changes requested):
can you update the JSIEngineOverride docs instead? (in the code repo, modify the doc_string in the IDL file for the API)

taenri:
I went ahead and proposed a change to the doc_string in the IDL file:
https://github.com/microsoft/react-native-windows/pull/8656

I think it would still be good to mention something about the JSIEngineOverride from Hermes.md documentation as well. That way other developers looking to experiment with Hermes will have some manner of discovering why hermes.dll might not be loading as expected when the UseHermes build flag is enabled. What do you all think?

jonthysell (approved):

jonthysell:
@asklar Does the updated text look okay?

Issue #527, opened 2021-09-01
Remove Certificates from Sample Apps


TatianaKapos:
The windows sample apps still have temporary certificates that need to be removed. Currently certificates are under publisher React Native Window Samples.

chrisglein:
@chiaramooney can you link the issue for removing it from the template?
In the meantime maybe we advise people to build debug when running samples so people aren't blocked when running this the first time (after cloning and trying to run samples).
Also if there's an issue tracking what we need to have release also work here, link that. Or create one.

chiaramooney:
Remove Cert from Template: https://github.com/microsoft/react-native-windows/issues/8202

TatianaKapos:
The only certificates left are in **Todosfeed and AppServiceDemo**. They both don't have workflows so the only work there would be to removing the certificate and updating the project file.

@qmatteoq Are you okay with me removing both certificates? I don't have much context for both apps so wanted to check in and make sure removing them won't break anything :)

PR #556, opened 2021-09-30
Add `--deploy-from-layout` to workflows

jonthysell:
In main/canary we've started removing certs from the templates, and
requiring that you use `--deploy-from-layout` when building with
`--release`. This PR preemptively makes these changes to our CI
workflows, so they'll keep building.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/556)

### DIFF
--- a/.github/workflows/samples-Calculator-CI.yml
+++ b/.github/workflows/samples-Calculator-CI.yml
@@ -38,7 +38,7 @@ jobs:
         working-directory: ..\..\src
 
       - name: Run react-native run-windows
-        run: npx react-native run-windows --logging --no-packager --no-launch --arch ${{ matrix.architecture }} ${{ startsWith(matrix.architecture, 'ARM') && '--no-deploy' || '' }} ${{ matrix.configuration == 'Release' && '--release --deploy-from-layout' || '' }}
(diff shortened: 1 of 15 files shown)


jonthysell:
Prep for #527

asklar:
@jonthysell  Should deploying from layout be the default for Release too if it's more necessary now, and when people are ready to build a package they can pass a new option?

NickGerleman:
+1, if we're not able to add logic to check if a project has AppX signing disabled, we should default deploy from layout to be true instead of false.

NickGerleman:
I'd be tempted to just change the default now regardless, and replace that with better behavior later when the msbuild evaluation work lands.

jonthysell:
> I'd be tempted to just change the default now regardless, and replace that with better behavior later when the msbuild evaluation work lands.

Creating a bug for the CLI here: https://github.com/microsoft/react-native-windows/issues/8773

In the meantime I'm going to set the `--deploy-from-layout` flag in these CIs because regardless of what direction we take, I don't think we're going to add certs back to this repo.

jonthysell:
Closing this PR so we can fix the CLI bug and see that it works here.

PR #573, opened 2021-10-26
Add specs for sample native modules

acoates-ms:
Fix a couple of issues with the current native module samples.

On the JS side they were exporting `undefined`.  Now they are exporting an object containing the two NativeModules that they implement.  The native modules are also now exporting JS type information.  This is flow for now, since this is what the packages are using currently.  This will also be able to be used with the codegen package in the future to generate turbomodules.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/573)


samples/NativeModuleSample/cppwinrt/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "native-module-sample",
+  "name": "native-module-sample-cppwinrt",
```
jonthysell, line 2: If you rename this, when following the upgrade steps in the readme, the solution file, vcxproj project file, and namespaces will all be different.

samples/NativeModuleSample/csharp/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "native-module-sample",
+  "name": "native-module-sample-csharp",
```
jonthysell, line 2: Same as for the cppwinrt version.

jonthysell:
Updated to latest main to remove spurious changes.

jonthysell (approved):

PR #577, opened 2021-10-29
Initial Doc for Upgrading RNW Apps

### LINKED ISSUE #533: Add Updating RNW App to New RNW Version Instructions
Do we have a page/section devoted to updating RNW apps on the website? If not, I think this would be a good addition. If we do and I missed it, is there a change we could make to have this section be more visible? 

I also just investigated the rn-upgrade-helper tool which now has support for Windows. Including information about this tool could also be helpful in that section.

chiaramooney:
**_Why_**
Currently, our website does not contain the steps to upgrade an app to a new version of RNW. 

Resolve #533 

**_What_**
Create an initial doc with steps for upgrade. Feel free to add input here on additional contents that would be helpful on this page. Would ultimately be good to add a troubleshooting section to this page with common errors during this process. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/577)


NickGerleman:
Upgrade helper is pretty fundamental to this. I think we should give it a callout.

docs/upgrade-app.md
```diff
@@ -0,0 +1,13 @@
...
+title: Upgrading App to Latest Version of React Native Windows
+---
+
+## Steps to Upgrade An App to New React Native Windows Version
+1. In the root of your project,  run `npx react-native-windows-init --version 0.XX.X --overwrite`.
+2. In the root of your project, run `npx react-native autolink-windows` to link any native modules used by your app.
+3. Step 1 will overwrite any changes you previously made to the native Windows code of your project. If you made changes to this code, you have two options:
+
+    1. Re-add those updates manually.
+    2. If your project is a GitHub repository, open the project in Visual Studio Code and navigate to the Source Control tab. There, you can undo the specific overwrite edits that erased your changes.
```
NickGerleman, line 12: nit: Git

docs/upgrade-app.md
```diff
@@ -0,0 +1,13 @@
+---
+id: upgrade-app
+title: Upgrading App to Latest Version of React Native Windows
+---
+
+## Steps to Upgrade An App to New React Native Windows Version
+1. In the root of your project,  run `npx react-native-windows-init --version 0.XX.X --overwrite`.
```
NickGerleman, line 7: Note that this will not remove any files that were removed from the new template

NickGerleman: This will also stamp over the project GUID, which we probably would like to avoid

chiaramooney: Sounds good will add these callouts!

asklar: I thought this was going to point to using upgrade helper? we should encourage automation rather than manually upgrading wherever possible (since it's error prone)

chiaramooney: Maybe I'm missing something here, but isn't upgrade helper the manual option? And then the RNW CLI is the automated option?

asklar: I see - I think what's missing here is that it doesn't call out upgrading via the CLI, it starts with "run react-native-windows-init" so I think we need a bulletpoint or two before that describes the automated process?
also, if we have the data in upgrade helper, why wouldn't react-native upgrade remove files that were removed from the template?

NickGerleman: `react-native upgrade` (the automated one) is no longer working/maintained. Upgrade helper, and the manual diffs it shows, is the current solution.

asklar: that's worrisome... the RN docs don't mention that: https://reactnative.dev/docs/upgrading
is there a bug tracking this / can we follow up?

chrisglein: I was about to ask why there wasn't a back link to the RN upgrading docs. I knew that Windows didn't support the `upgrade` command, but based on the docs I assumed it worked for other platforms. If it's no longer supported definitely open an issue and maybe just post a PR the core docs to delete that [whole section](https://reactnative.dev/docs/upgrading#react-native-cli).

Even if the `upgrade` command part is wrong, I still think we should have a link back to [Upgrading to newer versions](https://reactnative.dev/docs/upgrading), as it has additional details about upgrade helper and such.

chiaramooney: I have a direct link to the upgrade helper section of the react native docs on this page. I can add a link to the upgrading page in general, but it might be confusing to the reader since there is upgrade info on that page that isn't applicable to RNW.

chrisglein (approved):

PR #584, opened 2021-11-01
Move native API docs to their own folder for easier updating

asklar:
This will allow us to make release cutting simpler by simply deleting all of the contents in the native-api folder and copying the new docs over, rather than needing to merge them with the JS docs and figure out which docs were supposed to be deleted.

- [X] validated that links can be followed
- [x] validate the release cutting scripts work ok

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/584)


NickGerleman (approved):

chiaramooney (approved):
I believe release cutting scripts should be good to go since docusaurus supports this directory structure inside the docs directory and we only use scripts from docusaurus to cut the release

jonthysell:
You might want to check by locally "snapping" a new version, seeing what files get created, and verify running the site locally that the pages show up as you expect in the browser.

PR #587, opened 2021-11-04
Emphasize `noexcept` is a requirement #580

### LINKED ISSUE #580: Suggestion: emphasize `noexcept` is a must for `REACT_METHOD` and friends
### Page url

https://microsoft.github.io/react-native-windows/docs/native-modules

### Problem Description

The doc shows all examples with `noexcept`, but doesn't indicate it's a requirement. When I just started writing code for the native module, I decided I didn't want to handle exception requirement yet so I omitted `noexcept` in my method. I was then greeted with a wall of compile errors:
```
error C2027: use of undefined type 'winrt::Microsoft::ReactNative::ModuleMethodInfo<TMethod>'
with
[
TMethod=void (__cdecl winrt::namespace::Module::* )(winrt::Microsoft::ReactNative::ReactPromise<winrt::Microsoft::ReactNative::JSValue> &&)
]
```

Apparently `noexcept` is a [part of function type](https://en.cppreference.com/w/cpp/language/noexcept_spec) and `ModuleMethodInfo` has `noexcept` in its declaration so the compiler couldn't find the appropriate type when the method's specifier is `noexcept(false)` (i.e. omitted).

# Suggested fix
```
The REACT_MODULE macro-attribute says that the class is a React Native native module. It receives the class name as a first parameter. All other macro-attributes also receive their target as a first parameter. REACT_MODULE has an optional parameter for the module name visible to JavaScript and optionally the name of a registered event emitter. By default, the name visible to JavaScript is the same as the class name, and the default event emitter is RCTDeviceEventEmitter.

+Methods annotated with `REACT_METHOD` and friends must have the `noexcept` specifier. Module authors should make sure all exceptions are handled inside the method.

You can overwrite the JavaScript module name like this: REACT_MODULE(FancyMath, L"math").

You can specify a different event emitter like this: REACT_MODULE(FancyMath, L"math", L"mathEmitter").
```

Wording can be adjusted, but it is the suggested location for the emphasis.

roxk:
### Why

Resolves #580

### DIFF
--- a/docs/native-modules.md
+++ b/docs/native-modules.md
@@ -218,6 +218,8 @@ namespace NativeModuleSample
 
 The `REACT_MODULE` macro-attribute says that the class is a React Native native module. It receives the class name as a first parameter. All other macro-attributes also receive their target as a first parameter. `REACT_MODULE` has an optional parameter for the module name visible to JavaScript and optionally the name of a registered event emitter. By default, the name visible to JavaScript is the same as the class name, and the default event emitter is `RCTDeviceEventEmitter`.
 
+Methods annotated with `REACT_METHOD` and friends must have the `noexcept` specifier, otherwise the program will not compile. Module authors should make sure all exceptions are handled inside the method.
+
 You can overwrite the JavaScript module name like this: `REACT_MODULE(FancyMath, L"math")`.
 
 You can specify a different event emitter like this: `REACT_MODULE(FancyMath, L"math", L"mathEmitter")`.


docs/native-modules.md
```diff
@@ -218,6 +218,8 @@ namespace NativeModuleSample
 
 The `REACT_MODULE` macro-attribute says that the class is a React Native native module. It receives the class name as a first parameter. All other macro-attributes also receive their target as a first parameter. `REACT_MODULE` has an optional parameter for the module name visible to JavaScript and optionally the name of a registered event emitter. By default, the name visible to JavaScript is the same as the class name, and the default event emitter is `RCTDeviceEventEmitter`.
 
+Methods annotated with `REACT_METHOD` and friends must have the `noexcept` specifier, otherwise the program will not compile. Module authors should make sure all exceptions are handled inside the method.
```
jonthysell, line 221: Can you make sure this addtion is added to the older doc files?

I.e. under `website/versioned_docs/version-x.xx/nativemodules.md`

jonthysell:
Closed in favor of https://github.com/microsoft/react-native-windows-samples/pull/658

PR #601, opened 2021-11-17
Remove broken icon

asklar:
1. Removes broken icon in resources page:
Before:
image
After:
image

2. Prevents unnatural stretch of xbox img in showcase
Before:
image

After:
image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/601)

### DIFF
--- a/website/pages/en/resources-showcase.js
+++ b/website/pages/en/resources-showcase.js
@@ -27,13 +27,16 @@ function getBackgroundForCategory(category) {
 }
 
 function renderImg(img, isSmall, key) {
-  return <img src={img}
-    style={{
-      maxHeight: isSmall ? 40 : 500,
-      minHeight: isSmall ? 40 : 220,
-      minWidth: isSmall ? 40 : 220,
-      verticalAlign: 'end',
-    }} key={key} />;
+  return  <div
+  style={{
+    maxHeight: isSmall ? 40 : 600,
+    minHeight: isSmall ? 40 : 320,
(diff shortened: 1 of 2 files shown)


chrisglein:
> Removes broken icon in resources page

Screenshot of before/after?

website/pages/en/resources-showcase.js
```diff
@@ -27,13 +27,16 @@ function getBackgroundForCategory(category) {
...
+  return  <div
+  style={{
+    maxHeight: isSmall ? 40 : 600,
+    minHeight: isSmall ? 40 : 320,
+    minWidth: isSmall ? 40 : 320,
+    verticalAlign: 'baseline',
+  }}
+  ><img src={img}
+ key={key} />
+    </div>;
```
chrisglein, line 39: At this point you could probably scrap `img` and just set `background-image` on that `div`. Probably.

asklar: I don't think that does the same thing. Here's what I get if I replace img with ```backgroundImg: `url(${img})` ```

image

asklar:
> > Removes broken icon in resources page
> 
> Screenshot of before/after?

added to description

rectified95 (approved):
Approving. What was the faulty icon? Do we want to replace it with something similar?

PR #607, opened 2021-12-03
Add CLI Options Information to Website

### LINKED ISSUE #442: Document CLI options on the website
We have docs for the react-native-windows-init and for autolinking, but not for the CLI itself. 
We do have a [README.md file in the CLI](https://github.com/microsoft/react-native-windows/blob/master/packages/%40react-native-windows/cli/README.md) from where to source the information.

chiaramooney:
### Why
Add CLI options information to the website. This information currently only exists in the @react-native-windows/cli README. 

Resolves #442 
[#6620](https://github.com/microsoft/react-native-windows/issues/6620)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/607)


docs/getting-started.md
```diff
@@ -65,11 +65,39 @@ This sends telemetry to Microsoft by default. You can prevent the telemetry from
...
+| `--no-deploy`         | boolean    | Do not deploy the app                            |
+| `--deploy-from-layout`| boolean    | Force deploy from layout, even in release builds |
+| `--sln`               | string     | Override the app solution file determined by 'react-native config', e.g. windows\myApp.sln |
+| `--proj`              | string     | Override the app project file determined by 'react-native config', e.g. windows\myApp\myApp.vcxproj |
+| `--msbuildprops`      | string     | Comma separated props to pass to MSBuild, eg: prop1=value1,prop2=value2 |
+| `--buildLogDirectory` | string     | Optional directory where MSBuild log files should be stored |
+| `--info`              | boolean    | Dump environment information                     |
+| `--direct-debugging`  | number     | Enable direct debugging on specified port        |
+| `--useWinUI3`         | boolean    | Targets WinUI 3.0 (Preview) instead of UWP XAML. |
+| `--useHermes`         | boolean    | Use Hermes instead of Chakra as the JS engine (supported on 0.64+) |
```
AgneLukoseviciute, line 94:
```suggestion

```
Those two shouldn't be there. Were added to the wrong README in https://github.com/microsoft/react-native-windows/pull/7788, there's a fix in one of the open PRs just hasn't been merged in yet :)

chiaramooney: Good catch thanks!

docs/getting-started.md
```diff
@@ -65,11 +65,39 @@ This sends telemetry to Microsoft by default. You can prevent the telemetry from
...
+| `--no-telemetry`      | boolean    | Disables sending telemetry that allows analysis of usage and failures of the react-native-windows CLI |
+| `-h`, `--help`        | boolean    | output usage information                         |
+
   A new Command Prompt window will open with the React packager as well as a `react-native-windows` app. This step may take a while during first run since it involves building the entire project and all dependencies. You can now start developing! :tada:
 
-  This sends telemetry to Microsoft by default. You can prevent the telemetry from being sent by using the `--no-telemetry` command line option. See the `@react-native-windows/cli` README for more details.
-  
-  For a description of the options that `run-windows` offers, see [run-windows usage](https://github.com/microsoft/react-native-windows/blob/main/packages/@react-native-windows/cli/README.md#usage).
```
chrisglein, line 72: The plan is we keep these both up to date in duplicate? Do we have comments in the code to help guide people to what they need to update when they add/remove/change a CLI option? Is there anything more we expect the CLI readme will have that would justify keeping a back link here, or is it meant to be a subset?

chiaramooney: The CLI readme has to continue to exist because when the @react-native-windows/cli package is published to npm the README in that directory is used as the package description in npm. Information on CLI options is consistent across the README and what I'm adding to the website, only different is that the README has a longer blurb about the telemetry the package captures.

jonthysell: While I like moving some of this into the docs, I think it makes more sense to create a new page and not dump this giant table in the Getting Started doc.

jonthysell: Yeah, that table dump for init options would probably be better put a link away, i.e. "For more details on the options for react-native-windows-init, see XYZ"

chiaramooney: @jonthysell @chrisglein Is there a preference for how we want to organize the info? Should the link just take people back to the README or to the NPM page that contains the README info and well as versioning info, etc., or should it take people to another page on the website with the options laid out?

jonthysell: Because the CLI options change between versions, the link should go to a page with the correct table for that version. That could be the right release readme on NPM, though it might be easier to just have those pages within our docs and its versioning system.

As for the readme's themselves, as the CLI commands aren't really "standalone" (you only use them within the context of creating/building a RNW project) I'd be fine with reducing the readme to more-or-less a link back to our docs. That's what the normal CLI React Native does:

image

chiaramooney: Ok @jonthysell just update PR with those changes whenever you have a chance to review.

website/versioned_docs/version-0.63/cli.md
```diff
@@ -0,0 +1,77 @@
...
+| Option          | Input Type                                  | Description                                      |
+|-----------------|---------------------------------------------|--------------------------------------------------|
+| `--help`        | boolean                                     | Show help.                                       |
+| `--version`     | string                                      | The version of react-native-windows to use.      |
+| `--namespace`   | string                                      | The native project namespace.                    | 
+| `--verbose`     | boolean                                     | Enables logging.                                 |
+| `--language`    | string ["`cs`","`cpp`"] [default: "`cpp`"]  | Which language the app is written in.            |
+| `--projectType` | string ["`app`","`lib`"] [default: "`app`"] | The type of project to initialize.               |
+| `--overwrite`   | boolean                                     | Overwrite any existing files without prompting.  |
+| `--useWinUI3`   | boolean                                     | Targets WinUI 3.0 (Preview) instead of UWP XAML. |
```
asklar, line 31: is this content accurate? did we have winui3-ish support back in 0.63?
I recommend not having these pages in older releases. Only update 0.65+, the rest of the releases are not officially supported anymore as of next week

website/versioned_docs/version-0.63/cli.md
```diff
@@ -0,0 +1,77 @@
...
+title: React Native Windows CLI
+original_id: cli
+---
+
+This guide will give you more information on the React Native Windows CLI.
+
+## `react-native-windows-init`
+
+`react-native-windows-init` is a CLI used to bootstrap the addition of the Windows platform to `react-native` projects.
+
```
asklar, line 12: the challenge here is the init CLI is evergreen (there is only one), and the build CLI is per-version. So if we change the init cli, now we have to go update N copies of cli.md on the website.

Can this file be split into a init CLI and a per-version-build CLI file, and the init CLI doesn't go into versioned docs?
I mentioned to @jonthysell we should probably have some docs that don't get replicated for every version, he had some thoughts here too

website/versioned_docs/version-0.63/cli.md
```diff
@@ -0,0 +1,77 @@
...
+| `--proj`              | string     | Override the app project file determined by 'react-native config', e.g. `windows\myApp\myApp.vcxproj` |
+| `--msbuildprops`      | string     | Comma separated props to pass to MSBuild, e.g. `prop1=value1,prop2=value2` |
+| `--buildLogDirectory` | string     | Optional directory where MSBuild log files should be stored |
+| `--info`              | boolean    | Dump environment information                     |
+| `--direct-debugging`  | number     | Enable direct debugging on specified port        |
+| `--no-telemetry`      | boolean    | Disables sending telemetry that allows analysis of usage and failures of the react-native-windows CLI |
+| `-h`, `--help`        | boolean    | output usage information                         |
+
+This sends telemetry to Microsoft by default. You can prevent the telemetry from being sent by using the `--no-telemetry` command line option. See the `@react-native-windows/cli` [README](https://github.com/microsoft/react-native-windows/edit/main/packages/@react-native-windows/cli/README.md) for more details.
```
asklar, line 76: this link is to the "edit" page on github for the readme.md file, recommend either not linking to it, linking to the raw file, or to the rendered file instead

website/versioned_docs/version-0.63/cli.md
chiaramooney: Will do! That seems reasonable to me.

website/versioned_docs/version-0.63/cli.md
chiaramooney: Yea that definitely makes sense. I'm not sure if we fully figured out yet how to do static pages yet, but I can look into it! Currently we have the init CLI options written out on every getting-started page for every version, so we've already dug ourselves a bit of a whole with the N copies but it definitely seems like the best path would be to do the investment in setting up static pges.

website/versioned_docs/version-0.63/cli.md
chiaramooney: Will do!

website/versioned_docs/version-0.63/cli.md
jonthysell: Yeah, static pages are located in the website/pages folder. Unfortunately you have to write them as like actual ReactJS pages (there are examples in there), but I think it makes sense for the init command. Then we can just link to it from getting started, and from init's readme.

The full doc for `run-windows` and it's arguments should be versioned, and maybe worth renaming the doc from the generic cli.md to something like cli-run-windows.md.

chiaramooney: Sounds good!

jonthysell (approved):

PR #611, opened 2021-12-08
API Parity Page Rename

AgneLukoseviciute:
### Why
I think the term "API Parity" is not super intuitive for people outside the team/Microsoft. Renaming the tab to include "Components and APIs" might make more sense given the content of that page and since we rely on the RN Core documentation linked there for our non-windows-specific API documentation. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/611)

### DIFF
--- a/docs/parity-status.md
+++ b/docs/parity-status.md
@@ -1,6 +1,6 @@
 ---
 id: parity-status
-title: API Parity
+title: React Native Windows Components and APIs
 ---
 
 ## Core APIs and Components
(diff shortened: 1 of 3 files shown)


chiaramooney (commented):
Looks good to me!

chrisglein (approved):

jonthysell (approved):

PR #614, opened 2021-12-20
Update FancyMath.cs

Bowens20832:
## Description
Namespace is written twice. This would not compile if copied to Visual Studio.

### Why
Typo in the documentation.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/614)


jonthysell (approved):
Wow I can't believe it's been like that for two years. And that it's been building fine that whole time.

PR #616, opened 2022-01-05
(macOS) Project name error to 0.64

HeyImChris:
## Description
We had this error in our 0.63 docs but at some point it got omitted in our 0.64 docs. This change adds it back.

### Why
Someone in the OSS community reached out having hit this error. We should at least cover it in our documentation for now until a proper fix to the CLI tools is made.

Resolves [ProjectName not found error]

I validated it shows up as expected on my localhost.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/616)

### DIFF
--- a/website/versioned_docs/version-0.64/rnm-getting-started.md
+++ b/website/versioned_docs/version-0.64/rnm-getting-started.md
@@ -49,3 +49,6 @@ npx react-native-macos-init
   Open macos\test.xcworkspace in Xcode or run `xed -b macos`; `yarn start:macos`. Hit the Run button.
 
 A new Command Prompt window will open with the React packager as well as a `react-native-macos` app. This step may take a while during first run since it involves building the entire project and all dependencies. You can now start developing! 🎉
+
+### `ProjectName` has not been registered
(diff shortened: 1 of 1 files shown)


asklar (approved):

HeyImChris:
@asklar The "Check for broken links" CI timed out. Is that normal? With my permissions, I couldn't find a way to retrigger the CI run in case it's just an intermittent network issue causing the failure.

My change itself is pretty minor so I can't imagine that's causing a problem unless there's some other script I need to run too?

asklar:
> @asklar Alexander Sklar FTE The "Check for broken links" CI timed out. Is that normal? With my permissions, I couldn't find a way to retrigger the CI run in case it's just an intermittent network issue causing the failure.
> 
> My change itself is pretty minor so I can't imagine that's causing a problem unless there's some other script I need to run too?

nope you're good :) It was a transient error, I triggered a rerun and it passed, it usually takes <20 min to run PR build, but the last run got killed after 6h (the max time allowed).

PR #619, opened 2022-01-08
Add react-native-xaml to list of supported modules

asklar:
## Description
Adds react-native-xaml to the list of supported CMs.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/619)

### DIFF
--- a/docs/supported-community-modules.md
+++ b/docs/supported-community-modules.md
@@ -57,6 +57,7 @@ In addition to the ejected modules above, Microsoft has also added implementatio
 | <ins>[`react-native-code-push`](https://github.com/Microsoft/react-native-code-push)</ins> | 0.63 | x |
 | <ins>[`react-native-tts`](https://github.com/ak1394/react-native-tts)</ins> | 0.62 | x |
 | <ins>[`react-native-sketch-canvas`](https://github.com/creambyemute/react-native-sketch-canvas)</ins> | 0.63 | x |
+| <ins>[`react-native-xaml`](https://github.com/microsoft/react-native-xaml)</ins> | 0.64+ | x |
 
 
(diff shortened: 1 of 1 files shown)


chrisglein:
In the context of this comment above the list, react-native-xaml is a bit of an odd one out: 
```In addition to the ejected modules above, Microsoft has also added implementations for a set of popular and highly requested community modules```

Really, it's a Windows-specific module. Should we have a separate list of those? It could create room for the Glyphs and Flyouts of the world.

Also, that list in the docs is not necessarily comprehensive. We really want to keep [reactnative.directory](https://reactnative.directory/) up to date and direct people there, and probably deprecate this list. Or at least we've talked about it that way. What do you think?

jonthysell:
I agree that there should be some effort to push people toward the (up-to-date) directory, and maybe only list "highlights" here? Like, here are some top modules that we are engaged with - some were ejected from core, some were always community, and some are windows-only.

React-native-xaml definitely needs to be on that highlights list, if not also pointed to from elsewhere in the docs. Like, "looking for more xaml flexibility, why not try react-native-xaml?"

Having a table that shows a list of modules that we supported <= 0.63 feels like it's just going to cause confusion if those modules don't even work anymore.

jonthysell (approved):

chrisglein:
> React-native-xaml definitely needs to be on that highlights list, if not also pointed to from elsewhere in the docs. Like, "looking for more xaml flexibility, why not try react-native-xaml?"

Agreed that react-native-xaml should make the highlights list. Over time perhaps that's what this list becomes - the modules that we can vouch for some minimum bar of support (@stmoy  FYI).

PR #628, opened 2022-01-23
Document mixed `PackageReference` and `packages.config` projects

### LINKED ISSUE #621: Update documentation on using mixed PackageReference and packages.config dependencies.
Projects using `PackageReference` that have source dependencies on projects using `packages.config` may need to declare such sources' NuGet dependencies in a `packages.config` file of their own.

Example. For project structure:
```
MySolution\
|- MyMainProject\
   |- MyMainProject.vcxproj // Uses PackageReference
|- MyDependencyProject\
   |- MyDependencyProject.vcxproj
   |- packages.config
```

If `MyMainProject` has a source dependency on `MyDependencyProject`, because of the common NuGet/packages.config convention of placing NuGet packages inside directory `$(SolutionDir)packages`, `MyDependencyProject` may fail to build due to `msbuild.exe /t:restore` on `MyMainProject` extracting all dependencies into the NuGet variable `globalPackagesFolder`.

To make these hybrid dependency mechanism work, declare traditional NuGet packages in a `packages.config` file inside the `MyMainProject`'s directory.

This will create the `$(SolutionDir)packages` folder for any old-style dependencies introduced by the source dependency, while importing the rest from `globalPackagesFolder` (`PackageReference` mechanism).

_Originally posted by @NickGerleman in https://github.com/microsoft/react-native-windows/issues/9328#issuecomment-1007844969_

JunielKatarn:
## Description

See https://github.com/microsoft/react-native-windows/pull/8195

Adds entry with guidance for mixed `PackageReference`/`packages.config` projects.

### Why

After introducing `PackageReference` as MSRN's native NuGet dependency mechanism, source dependencies that use `packages.config` will still expect their own NuGet deps to be located in `$(SolutionDir)packages\`.

Resolves #621

### DIFF
--- a/docs/managing-cpp-deps.md
+++ b/docs/managing-cpp-deps.md
@@ -1,0 +1,48 @@
+---
+id: managing-cpp-deps
+title: Managing C++ dependencies
+---
+
+Details to consider when consuming a Visual C++ project as a source dependency of your app.
+
+### Details
+
+Applications with native code, either written on `C#` or `C++`, may add source dependencies on native Visual C++ (`.vcxproj`) projects.\
+Meaning, such dependencies will be built as part of the application.
+
+Starting with version `0.68`, React Native for Windows apps use the [`PackageReference`](https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files) restore project style for native `C++` NuGet dependencies.\
+The main change consists in NuGet packages being directly loaded from the user account's `globalPackagesFolder` cache instead of also copying them into a local folder relative to the Visual Studio Solution location (see [repositoryPath](https://docs.microsoft.com/en-us/nuget/reference/nuget-config-file)).
+
+This may conflict with Node/Visual C++ dependencies that use the more common [`packages.config`](https://docs.microsoft.com/en-us/nuget/reference/packages-config) project style (i.e. [React Native Picker](https://github.com/react-native-picker/picker#react-native-pickerpicker)).
+
+## Updating your app
+
+### Add the React Native Windows C++ Dependency (i.e. RN Picker)
+
+```
+yarn add @react-native-picker/picker
+```
+
+### Add a `packages.config` file to your app's project directory
+
+The React Native Windows build system will try to build the dependency from source.\
+Because it uses the `packages.config` restore style, it will most likely expect its own NuGet dependencies to be restored at `$(SolutionDir)packages\`.
+
+#### See [node_modules\@react-native-picker\picker\windows\ReactNativePicker\ReactNativePicker.vcxproj](https://github.com/react-native-picker/picker/blob/v2.2.1/windows/ReactNativePicker/ReactNativePicker.vcxproj#L156)
+
+```xml title="ReactNativePicker.vcxproj"
+<Import Project="$(SolutionDir)\packages\Microsoft.Windows.CppWinRT.2.0.210312.4\build\native\Microsoft.Windows.CppWinRT.targets" Condition="Exists('$(SolutionDir)\packages\Microsoft.Windows.CppWinRT.2.0.210312.4\build\native\Microsoft.Windows.CppWinRT.targets')" />
+<Import Project="$(SolutionDir)\packages\$(WinUIPackageName).$(WinUIPackageVersion)\build\native\$(WinUIPackageName).targets" Condition="Exists('$(SolutionDir)\packages\$(WinUIPackageName).$(WinUIPackageVersion)\build\native\$(WinUIPackageName).targets')" />
+```
+
+#### `windows\<C++ or C# app directory>\packages.config`
+
+```xml title="packages.config"
+<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="Microsoft.Windows.CppWinRT" version="2.0.210312.4" targetFramework="native" />
+  <package id="Microsoft.UI.Xaml" version="2.3.191129002" targetFramework="native" />
+</packages>
+```
+
+This way, your React Native Windows app will restore the external NuGet dependencies into the expected location.
--- a/docs/customizing-SDK-versions.md
+++ b/docs/customizing-SDK-versions.md
@@ -46,8 +46,6 @@ When you create a C# or C++ React Native for Windows app, it is written in such 
 
 For C++ apps, in addition to setting the property value, you will also need to modify the `packages.config` file which can be found next to the app project `.vcxproj` file, to point to the updated package version.
 
-<!--DOCUSAURUS_CODE_TABS-->
-<!--C# app-->
 
 #### `windows\ExperimentalFlags.props`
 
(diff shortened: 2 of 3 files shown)


NickGerleman:
@asklar is Unbroken respectful to rate limiting requests that services give? It looks like we're seeing live network failure when accessing Github in the markdown link checking script.

docs/managing-cpp-deps.md
```diff
@@ -0,0 +1,48 @@
+---
+id: managing-cpp-deps
+title: Managing C++ dependencies
+---
+
+Details to consider when consuming a Visual C++ project as a source dependency of your app.
```
NickGerleman, line 6: A major part of the RNW demographic will be familiar with Visual Studio project structure. From what I've observed, others have expertise in JS, and rely on tools like "autolinking" to manage dependencies for them.

That audience may not know, that under the hood, their npm package dependency brings in a vcxproj. The term the community would recognize for these npm dependencies bringing in native code is "community modules". Since that has the potential to be such a common case, could we call them out by name throughout the document?

E.g.

```suggestion
Details to consider when consuming community modules or other Visual C++ projects.
```

docs/managing-cpp-deps.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Details
+
+Applications with native code, either written on `C#` or `C++`, may add source dependencies on native Visual C++ (`.vcxproj`) projects.\
+Meaning, such dependencies will be built as part of the application.
+
+Starting with version `0.68`, React Native for Windows apps use the [`PackageReference`](https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files) restore project style for native `C++` NuGet dependencies.\
+The main change consists in NuGet packages being directly loaded from the user account's `globalPackagesFolder` cache instead of also copying them into a local folder relative to the Visual Studio Solution location (see [`repositoryPath`](https://docs.microsoft.com/en-us/nuget/reference/nuget-config-file)).
+
+This may conflict with Node/Visual C++ dependencies that use the more common [`packages.config`](https://docs.microsoft.com/en-us/nuget/reference/packages-config) project style (i.e. [React Native Picker](https://github.com/react-native-picker/picker#react-native-pickerpicker)).
```
NickGerleman, line 16: There are many community modules, so the reader might be using a different one, and not be familiar with Picker. 

```suggestion
This may conflict with Visual C++ dependencies that use the more common [`packages.config`](https://docs.microsoft.com/en-us/nuget/reference/packages-config) project style, including community modules generated targeting versions before 0.68.
```

NickGerleman: Same feedback for below, for anywhere we reference anything specific to Picker.

docs/managing-cpp-deps.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Updating your app
+
+### Add the React Native Windows C++ Dependency (i.e. RN Picker)
+
+```
+yarn add @react-native-picker/picker
+```
+
+### Add a `packages.config` file to your app's project directory
```
NickGerleman, line 26: Historically, autolinking would manage this file. Is that still the case?

NickGerleman: Er, I might be mixing this up. I remember there was some differences of NuGet restore behavior for transitive dependencies on packages.config that @jonthysell ran into before.

NickGerleman:
Another case to document is raw MSBuild. I.e. we have special logic in run-windows, and hooks for VS, but those using MSBuild directly might need to do the second restore themselves.

JunielKatarn:
> Another case to document is raw MSBuild. I.e. we have special logic in run-windows, and hooks for VS, but those using MSBuild directly might need to do the second restore themselves.

Based on the original discussions around `PackageReference` back in July 2021, it was my understanding that we were not directly supporting raw MSBuild.
That said, I'm happy to add these use cases.

HOWEVER, as I mentioned offline, I may have found a way to make these docs unnecessary.
I'll poke around and have a clear answer early next week.

NickGerleman:
> > Another case to document is raw MSBuild. I.e. we have special logic in run-windows, and hooks for VS, but those using MSBuild directly might need to do the second restore themselves.
> 
> Based on the original discussions around `PackageReference` back in July 2021, it was my understanding that we were not directly supporting raw MSBuild. That said, I'm happy to add these use cases.
> 
> HOWEVER, as I mentioned offline, I may have found a way to make these docs unnecessary. I'll poke around and have a clear answer early next week.

I think we should explore the solution, but the 0.68 preview will go out on Monday. If possible, it would be nice to document the current state of the world, for anyone trying to test it.

docs/managing-cpp-deps.md
jonthysell: Wait, is this just as simple as adding a packages.config to that app's project directory?

Is there any reason we're not just putting/modifying a file there ourselves during autolinking? Or just manually calling nuget restore on the lib project so the files get put in the right place?

JunielKatarn: Calling NuGet restore doesn't work once PackageReference is detected.

I did not know we could have Autolink set this up for us.
We could add such functionality and then remove this instruction.

chiaramooney (commented):
Looks good to me @jonthysell @asklar Any final adjustments needed?

JunielKatarn:
> Looks good to me @jonthysell @asklar Any final adjustments needed?
> 
> @JunielKatarn If you rebase the broken webview links should be fixed.

I actually found a way that could make these steps, and ergo, this PR unnecessary.
Discussing offline.

(Keeping the PR open until a decision is made).

JunielKatarn:
> > I actually found a way that could make these steps, and ergo, this PR unnecessary. Discussing offline.
> > (Keeping the PR open until a decision is made).
> 
> @JunielKatarn Any updates on that?

@chrisglein for the time being, we are keeping this approach. The mentioned alternative would be invasive and we'd rather not take any further disruptive changes, at least for the 0.68 release.

I will update and merge this PR once approved.

docs/managing-cpp-deps.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Details
+
+Applications with native code, either written on `C#` or `C++`, may add source dependencies on native Visual C++ (`.vcxproj`) projects.\
+Meaning, such dependencies will be built as part of the application.
+
+Starting with version `0.68`, React Native for Windows apps use the [`PackageReference`](https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files) restore project style for native `C++` NuGet dependencies.\
+The main change consists in NuGet packages being directly loaded from the user account's `globalPackagesFolder` cache instead of also copying them into a local folder relative to the Visual Studio Solution location (see [`repositoryPath`](https://docs.microsoft.com/en-us/nuget/reference/nuget-config-file)).
+
+This may conflict with Visual C++ dependencies that use the more common [`packages.config`](https://docs.microsoft.com/en-us/nuget/reference/packages-config) project style, including community modules generated targeting versions before 0.68.
```
asklar, line 16:
```suggestion
Starting with version `0.68`, React Native for Windows apps use the [`PackageReference`](https://docs.microsoft.com/nuget/consume-packages/package-references-in-project-files) restore project style for native `C++` NuGet dependencies.\
The main change consists of NuGet packages being placed directly in the user account's `globalPackagesFolder` cache instead of also copying them into a local folder relative to the Visual Studio Solution location (see [`repositoryPath`](https://docs.microsoft.com/nuget/reference/nuget-config-file)).

This may conflict with C++ dependencies that use the more common [`packages.config`](https://docs.microsoft.com/nuget/reference/packages-config) project style, including community modules generated targeting versions older than 0.68.
```

docs/managing-cpp-deps.md
```diff
@@ -0,0 +1,67 @@
...
+Because it uses the `packages.config` restore style, it will most likely expect its own NuGet dependencies to be restored at `$(SolutionDir)packages\`.
+
+#### See [`node_modules\@react-native-picker\picker\windows\ReactNativePicker\ReactNativePicker.vcxproj`](https://github.com/react-native-picker/picker/blob/v2.2.1/windows/ReactNativePicker/ReactNativePicker.vcxproj#L156)
+
+```xml title="ReactNativePicker.vcxproj"
+<Import Project="$(SolutionDir)\packages\Microsoft.Windows.CppWinRT.2.0.210312.4\build\native\Microsoft.Windows.CppWinRT.targets" Condition="Exists('$(SolutionDir)\packages\Microsoft.Windows.CppWinRT.2.0.210312.4\build\native\Microsoft.Windows.CppWinRT.targets')" />
+<Import Project="$(SolutionDir)\packages\$(WinUIPackageName).$(WinUIPackageVersion)\build\native\$(WinUIPackageName).targets" Condition="Exists('$(SolutionDir)\packages\$(WinUIPackageName).$(WinUIPackageVersion)\build\native\$(WinUIPackageName).targets')" />
+```
+
+#### `windows\<C++ or C# app directory>\packages.config`
```
asklar, line 38: if the app is C#, would it ever use packages.config?

JunielKatarn: Not directly, but a C# project could depend on a C++ project that uses `packages.config`.

For instance, a MSRN C# app could depend on RN Picker, couldn't it?

chiaramooney (approved):
👍

PR #719, opened 2022-07-23
CoreApp blog post and sample

asklar:
## Description
Adds a blog post about the new CoreApp API

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/719)


website/blog/2022-07-25-coreapp.md
```diff
@@ -0,0 +1,128 @@
...
+
+We heard from some of our customers that build times for these types of apps can be pretty long - especially so when building the React Native for Windows framework from source. To mitigate this problem we offer pre-built binary distributions of the framework in the form of NuGet packages. This works pretty well but has a few shortcomings:
+
+1) Even when not building the framework, building a UWP requires several build steps behind the scenes (the steps below are what happens for C++/WinRT apps):
+    - the MIDL compiler is run to turn IDL files into Windows Metadata - this is required to produce information used by the XAML compiler
+    - the XAML compiler is invoked to convert XAML markup files into C++ code
+    - the MDMerge tool runs to merge all Windows Metadata files from the application
+    - C++/WinRT is invoked to turn Windows Metadata into C++ projection headers as well as component sources
+    - Only then, the C++ compiler and linker run on the app
+2) Native modules don't currently fully support consuming the RNW NuGet package that the app they are getting built into is using; there is work in progress to unlock this scenario.
```
stmoy, line 20: I'm not sure what this sentence is saying. Is it:
" Native modules can't consume the prebuilt RNW NuGet package when they are included in an app that depends on it (though there is work outstanding to fix this)"?

Would recommend linking to the tracking issue.

website/blog/2022-07-25-coreapp.md
```diff
@@ -0,0 +1,128 @@
...
+```cpp
+extern "C" __declspec(dllexport) void *__cdecl MySpecialPackageProvider() {
+  auto provider = winrt::make<MyModulePackageProvider>();
+  void *abi{nullptr};
+  winrt::copy_to_abi(provider, abi);
+  return abi;
+}
+```
+
+This snippet will declare a plain C function (this is important so that C++ name mangling mechanism doesn't come into play, which is required to be able to use the function name in the JSON file), and export it from the DLL (or the application's EXE). It takes no parameters, and simply return a pointer to the package provider for the module.
```
stmoy, line 107: The formatting from here out looks funky - might just be strangely rendered in GitHub, but probably want to double check that it looks good when published.

website/blog/2022-07-25-coreapp.md
```diff
@@ -0,0 +1,128 @@
...
+    - the XAML compiler is invoked to convert XAML markup files into C++ code
+    - the MDMerge tool runs to merge all Windows Metadata files from the application
+    - C++/WinRT is invoked to turn Windows Metadata into C++ projection headers as well as component sources
+    - Only then, the C++ compiler and linker run on the app
+2) Native modules don't currently fully support consuming the RNW NuGet package that the app they are getting built into is using; there is work in progress to unlock this scenario. 
+3) There's no framework source to patch if you want to customize some behavior; this is usually not needed by the vast majority of customers, but listing this here for completeness.
+
+We set out to address the first item on the list above, since a lot of applications won't need to customize their native code; these _greenfield_ applications consist of the bare-bones native UWP app, and all of their app logic is confined to their JavaScript (and whatever native modules they require).
+
+## Enter `CoreApp`
```
stmoy, line 25: Recommend adding a "get started" section. Is this going to be the new default mode for apps created in 0.70?

website/blog/2022-07-25-coreapp.md
```diff
@@ -0,0 +1,128 @@
...
+To solve this issue, we came up with a very lean, yet flexible, API that enables a UWP Core Application (the kind that doesn't require XAML, or any interface definition) to host these _greenfield_ apps. We call this API `CoreApp`.
+When using the `CoreApp` API, the React Native for Windows framework provides the required `XamlApplication` object as well as a simple XAML `Page` object, which contains a React root view.
+
+This is a C API that comes in two flavors:
```
stmoy, line 29: Will there be some official docs and/or samples in addition to the blog?

website/.unbroken_exclusions
```diff
@@ -1,5 +1,6 @@
 !node_modules
 !blog
+!versioned_docs/version-0.60/e2e-test-more-about.md
```
stmoy, line 3: Is this correct? Does there need to be a new one for the *-coreapp.md file?

stmoy (commented):
Thanks for writing this! Added some questions/comments below.

website/.unbroken_exclusions
asklar: no, blogs are not checked (see line number 2); 
this is needed because in the 0.60 e2e testing doc, we have a link to a wdio package and that package has been removed in latest versions of wdio, so the link is now broken. This gets our CI passing again.

website/blog/2022-07-25-coreapp.md
asklar: not default, I should clarify this is experimental / trying to get customer feedback from it.

website/blog/2022-07-25-coreapp.md
asklar: validated this looks ok on the website, lmk if you disagree:
image

website/blog/2022-07-25-coreapp.md
asklar: we don't currently have NuGet packages for canary so it'll be a little tricky for folks to build with NuGet. For the time being I will point folks to the coreapp sample app, and I'll update the blog post once this gets into a nuget (once 70 ships)

website/blog/2022-07-25-coreapp.md
asklar: actually this might not be true anymore after re-reading this comment:
https://github.com/microsoft/react-native-windows/issues/7227#issuecomment-1136441556

@jonthysell is that right? modules created in 0.68 don't need to do anything to use their app's RNW nuget (instead of building RNW from source - and they don't need to build RNW from source anymore?)

website/blog/2022-07-25-coreapp.md
asklar: added an API doc; this is the first Cxx style API we are documenting "officially".

In parallel, last year @vmoroz had looked into using docusaurus for producing docs for the other ones, is that ready to be used on the other C++-only APIs like ReactPackageProvider etc?

website/blog/2022-07-25-coreapp.md
asklar: removed that whole mention, assuming the comment about NuGet not working for modules is no longer accurate

website/blog/2022-07-25-coreapp.md
asklar: filed #720 to track using his package to create api docs for the other cxx apis

stmoy (approved):

.github/workflows/samples-CalculatorCoreAppNuGet-ci-upgrade.yml
```diff
@@ -0,0 +1,65 @@
...
+        run: yarn install
+        working-directory: ..\..\src
+
+      - name: Upgrade react-native-windows
+        id: upgrade
+        continue-on-error: true
+        run: ${{github.workspace}}\.github\scripts\UpgradeSmokeTest.ps1 ${{ matrix.reactNativeWindowsVersion }} -Force ${{ format('${0}', contains(matrix.reactNativeWindowsVersion, 'canary')) }}
+        working-directory: ..\..\src
+
+      - name: Run react-native-windows-init
```
jonthysell, line 42: I know you're already merged this, but you're gonna want to comment out the `react-native-windows-init` task here until you get the template created. Otherwise this is just going to fail.

docs/.unbroken_exclusions
```diff
@@ -0,0 +1 @@
+!coreapp.md
```
jonthysell, line 1: Why is this needed here?

PR #722, opened 2022-08-24
Cut Website Documentation for v0.70

chiaramooney:
### Why
Cut Website Documentation for v0.70

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/722)


jonthysell:
The support page doesn't have a line for 0.70. It should have been marked as Preview now, but obviously we're going through preview pretty fast. Do we want to include that line as part of this? Then when it's promoted to latest we can do another PR to change it to from preview to Active?

chiaramooney:
@jonthysell Sure! I'll add that in. I'll probably just add the line as Active to this PR because this will just be merged when 70 stable is released because of the tight timeline this time around.

jonthysell (approved):

PR #727, opened 2022-09-08
Rework native modules docs to use spec files

acoates-ms:
This updates the documentation for native modules to use JS/TS spec files, and be compatible with the native code being either a TurboModule or a Native Module.

Is also shows them how to use codegen to ensure that the definition in native aligns with the definition in JS/TS.  And provides information on how to make their modules run as turbomodules.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/727)


vmoroz (approved):
:shipit:

PR #781, opened 2023-02-17
New JavaScript Debugging Doc

jonthysell:
This PR adds a new doc outlining how to debug the JavaScript of your RNW application, including support matrices and step-by-step instructions for connecting all supported JS debuggers.

This doc was made by testing **every** combination against RNW 0.71 with newly created RNW projects. These steps *may* work for older RNW versions, but I haven't tested that (yet), so I won't backport the doc to earlier versions until I'm sure.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/781)

### DIFF
--- a/docs/debugging-javascript.md
+++ b/docs/debugging-javascript.md
@@ -122,7 +122,7 @@ You can web debug within [VS Code](http://code.visualstudio.com/) by using the [
     ```
     3. Ensure that Metro, your native app, and/or any browser Developer Tools are **not** already running
     4. Ensure the new config is selected at the top of the *Run and Debug* sidebar
-    5. Click on the ▶️ button or press *F5* in VS Code
+    5. Click on the ▶️ button or press `F5` in VS Code
     > Watch VS Code's *Debug Console* output for a successful debugger connection. This can be very slow, but VS Code should automatically launch Metro, establish the debugger connection, then launch your native app to download the bundle.
 6. **Option B:** Attach VS Code to Metro that's already running
     1. Click on the ⚙️ to the right of *Attach to packager*
@@ -138,7 +138,7 @@ You can web debug within [VS Code](http://code.visualstudio.com/) by using the [
     ```
     3. Ensure sure that only Metro is already running (i.e. `npx react-native start`)
     4. Ensure the new config is selected at the top of the *Run and Debug* sidebar
-    5. Click on the ▶️ button or press *F5* in VS Code
+    5. Click on the ▶️ button or press `F5` in VS Code
     6. Confirm the address of the Metro server (default: `localhost`)
     7. Confirm the port of the Metro server (default: `8081`)
     > Watch VS Code's *Debug Console* output for a successful debugger connection.
@@ -283,7 +283,7 @@ You can direct debug RNW apps using the Hermes JS engine with [VS Code](http://c
     ```
     3. Ensure that Metro is already running (i.e. `npx react-native start`)
     4. Ensure the new config is selected at the top of the *Run and Debug* sidebar
-    5. Click on the ▶️ button or press *F5* in VS Code
+    5. Click on the ▶️ button or press `F5` in VS Code
     6. Confirm the address of the Metro server (default: `localhost`)
     7. Confirm the port of the Metro server (default: `8081`)
     > Watch VS Code's *Debug Console* output for a successful debugger connection.
@@ -320,7 +320,7 @@ You can direct debug RNW apps using the Hermes JS engine with [VS Code](http://c
     ```
     3. Ensure that Metro and your native app are **not** already running
     4. Ensure the new config is selected at the top of the *Run and Debug* sidebar
-    5. Click on the ▶️ button or press *F5* in VS Code
+    5. Click on the ▶️ button or press `F5` in VS Code
     > Watch VS Code's *Debug Console* output for a successful debugger connection. This can be very slow, but VS Code should automatically launch Metro, establish the debugger connection, then launch your native app to download the bundle.
 6. **Option B:** Attach VS Code to Metro that's already running
     1. Click on the ⚙️ to the right of *Attach to Hermes application - Experimental*
@@ -336,7 +336,7 @@ You can direct debug RNW apps using the Hermes JS engine with [VS Code](http://c
     ```
     3. Ensure that Metro is already running (i.e. `npx react-native start`)
     4. Ensure the new config is selected at the top of the *Run and Debug* sidebar
-    5. Click on the ▶️ button or press *F5* in VS Code
+    5. Click on the ▶️ button or press `F5` in VS Code
     6. Confirm the address of the Metro server (default: `localhost`)
     7. Confirm the port of the Metro server (default: `8081`)
     > Watch VS Code's *Debug Console* output for a successful debugger connection.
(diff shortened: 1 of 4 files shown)


vmoroz:
@jonthysell , this is an awesome doc! Thank you for doing it!
Going forward we must drop support for Web Debugging.
It does not work with synchronous TurboModules and JSI. 
Could you add it to the docs and move the Direct Debugging on top of Web Debugging to promote it more.
I also wonder if we should add Edge/Chrome as an option for the Direct Debugging?

jonthysell:
Once the doc is signed off, I'll amend this PR to also backport it to 0.71 so it shows up for users right away.

docs/debugging-javascript.md
```diff
@@ -0,0 +1,382 @@
...
+
+## Web vs. Direct Debugging
+
+As mentioned above, RNW supports both Web and Direct Debugging, and it's worth knowing when to use each.
+
+Web Debugging is the older, more well-established of the two debugging solutions. It relies on the powerful in-browser debugging tools already familiar to web developers. This reinforces RN's purpose of enabling web developers to build native apps with web technologies.
+
+However, Web Debugging is not without its limitations. The primary problem is that JS can behave slightly differently when running in different JS engines. As such, by developing (and testing) your app while using a browser to run the JS, you might miss bugs or other issues that only manifest when running your app normally with the embedded JS engine.
+
+This puts extra onus on developers to test all of their app's functionality in their release build, or otherwise risks situations where customers find bugs in production that aren't reproducible when using Web Debugging.
```
vmoroz, line 360: >onus [](http://example.com/codeflow?start=16&length=4)

Could we use an alternative word? It seems to be a quite rare word which I personally see the first time. The docs should be readable not only to native English speakers.

docs/debugging-javascript.md
```diff
@@ -0,0 +1,382 @@
...
+    ```json
+    {
+        "name": "Attach to packager",
+        "cwd": "${workspaceFolder}",
+        "type": "reactnative",
+        "request": "attach",
+        "isDynamic": true
+    }
+    ```
+    3. Ensure sure that only Metro is already running (i.e. `npx react-native start`)
```
vmoroz, line 139: >Ensure sure [](http://example.com/codeflow?start=7&length=11)

"Make sure"?

docs/debugging-javascript.md
```diff
@@ -0,0 +1,382 @@
...
+    7. Confirm the port of the Metro server (default: `8081`)
+    > Watch VS Code's *Debug Console* output for a successful debugger connection.
+
+    8. Manually launch your native app
+
+Once everything is running, you should be able to debug your RNW application. To set breakpoints, simply do so in your app's JS source files directly in VS Code.
+
+> Remember to save `launch.json` so you can reuse this debugging configuration in the future.
+
+> *Important:* If you're having difficulty getting VS Code to connect with Web Debugging, make extra sure you're not running any part of RNW (watch out for rogue `node.exe` processes) before starting the steps above.
```
vmoroz, line 152: >make extra sure [](http://example.com/codeflow?start=90&length=15)

Re-phrase? E.g. "Double check"

docs/debugging-javascript.md
jonthysell, line 139:
```suggestion
    3. Make sure that only Metro is already running (i.e. `npx react-native start`)

```

docs/debugging-javascript.md
jonthysell, line 152:
```suggestion
> *Important:* If you're having difficulty getting VS Code to connect with Web Debugging, double check you're not running any part of RNW (watch out for rogue `node.exe` processes) before starting the steps above.

```

jonthysell:
@vmoroz:

> @vmoroz: Going forward we must drop support for Web Debugging. It does not work with synchronous TurboModules and JSI. Could you add it to the docs and move the Direct Debugging on top of Web Debugging to promote it more.

I went back and forth on how which to put first. I know the eventual plan is to drop web debugging, but as of 0.71 it's still the default solution for new RN and RNW apps. The official [RN Debugging page](https://reactnative.dev/docs/debugging) only talks about Web Debugging. It's only [buried in the Hermes docs](https://reactnative.dev/docs/hermes#debugging-js-on-hermes-using-google-chromes-devtools) that Direct Debugging is mentioned. Even our React Native Tools VS code extension still labels direct debugging for Hermes as "Experimental".

I do mention synchronous calls in the "Web vs. Direct" section, and that the plan is to invest more in direct debugging. But I don't think we should put Direct Debugging first until it's the default for new apps and Web Debugging is officially and explicitly marked as legacy/unsupported.

> @vmoroz: I also wonder if we should add Edge/Chrome as an option for the Direct Debugging?

I tried really hard to get Edge's "dedicated DevTools for Node" to work with direct debugging. I actually had the setup steps written in an earlier draft, but the tools were so unreliable. Sometimes it wouldn't see Hermes at all. Sometimes it would, but when you connected it just loaded an empty window with only the console log (no tools to actually debug).

At best, it would connect, and appear to load sources and everything properly, but setting breakpoints didn't actually work. Setting a breakpoint in say, a button's onPress callback, would instead cause the debugger to set a breakpoint buried in some module loading function that got hit on every JS thread tick. I was never able to get a basic debugging session working, so I decided to leave it out of the docs.

docs/debugging-javascript.md
```diff
@@ -0,0 +1,382 @@
...
+#### Using the Edge Developer Tools
+
+You can web debug within Edge by using the [Edge Developer Tools](https://learn.microsoft.com/en-us/microsoft-edge/devtools-guide-chromium/).
+
+> If you use Chrome, the process is similar for using the [Chrome Developer Tools](https://developer.chrome.com/docs/devtools/).
+
+1. Launch your RNW app (with Web Debugging enabled and Metro running)
+2. In Edge, open the URL http://localhost:8081/debugger-ui/
+    > Metro may have already opened it, look for a *React Native Debugger* tab.
+3. Press `Ctrl+Shift+J` on the page to launch the Developer Tools
```
chrisglein, line 93: I've been using Ctrl+Shift+I all this time. TIL that both I or J work.

docs/debugging-javascript.md
```diff
@@ -0,0 +1,382 @@
...
+| JavaScript Engine | Visual Studio | Visual Studio Code | Visual Studio Code<br/> w/ React Native Tools |
+|:------------------|:-:|:-:|:-:|
+| Chakra (Default)  | ✅ | 🟥 | 🟥 |
+| Hermes            | 🟥 | ✅ | ✅ |
+
+> **Important:** Direct Debugging is relatively new and may still have some rough edges, depending on your choice of engine and debugger. See [Web vs. Direct Debugging](#web-vs-direct-debugging) for details.
+
+### Step 1: Enable Direct Debugging
+
+You have two options to enable Direct Debugging: at compile-time in your app's native code or at runtime via the in-app Developer Menu.
```
chrisglein, line 175: Also from CLI, yeah? `npx react-native run-windows --direct-debugging`
https://github.com/microsoft/react-native-windows/commit/c1a2140928b37d9c761c41bfa0fcc38704a14099

chrisglein (approved):

docs/debugging-javascript.md
jonthysell: Yes but I had trouble getting it to work - I think there may be a bug where those settings get overwritten.

PR #796, opened 2023-03-10
Outdated typescript instructions

JakeHedman:
Remove reference to `react-native-template-typescript` template which has been deprecated since React Native added first-class support for Typescript in version 0.71
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/796)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/796)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/796)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/796)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/796)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/796)

PR #798, opened 2023-03-14
Update getting-started.md

An631:
Starting from React Native v0.71 TypeScript is used by default

## Description
Adds a quick comment regarding the use of typescript by default after version 0.71

### Why
The use of the typescript templates is no longer necessary and we still recommend it in the documentation without specifying when it is needed.

## Screenshots
image

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/798)

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -18,7 +18,7 @@ npx react-native init <projectName> --template react-native@^0.71.0
 ```
 
 > To create TypeScript template, run `npx react-native init <projectName> --template react-native-template-typescript`.<br><br>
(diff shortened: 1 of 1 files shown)


chrisglein:
Thanks for raising this!
Similar to https://github.com/microsoft/react-native-windows-samples/pull/796
Question is whether it's clearer to delete the reference to the template entirely. I think that's clearer? In which case that'd be the PR to merge.

PR #826, opened 2023-04-06
Add Release builds to PR checks and update AppServiceDemo to 0.71

jonthysell:
## Description

This PR updates the PR workflow to also build Release configurations, not just Debug. This PR also updates the AppServiceDemo to 0.71 because that's the only way to build its Release configuration in CI.

### Why

Since we can only build the apps in workflows, not actually run them, by only building Debug we only verify that the native code builds, not that the JS code can be bundled, loaded, and run. By building Release, we get the automatic app bundling, at least confirming that the JS code can be bundled without errors.

In attempting this, it was determined that the AppServiceDemo could not build in Release because that bundling would fail. Specifically, it gets error `error:0308010C:digital envelope routines::unsupported`, because AppServiceDemo targets the (very) old RNW 0.67 and the `windows-2019` image used in CI uses Node 18, which causes the error.

One solution would be to install the older Node 16 onto the machines, but that sort-of defeats the purpose of the samples as verification that you can build RNW apps using GitHub Actions and the public VM images. The other option would be to upgrade AppServiceDemo to RNW 0.71 alongside all the other samples. This required a little more work (the sample itself is a non-standard use of RNW) but ultimately brings all samples in line with one another without needing "special" workflows for each.

## Screenshots
N/A

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/826)


jonthysell:
FWIW, upgrading the AppServiceDemo revealed a regression in RNW 0.68, which is being addressed here: https://github.com/microsoft/react-native-windows/pull/11464

When that PR is merged and backported, it'll be worth bumping AppServiceDemo again, as this will help improve its build reliability on local machines.

chrisglein (approved):

PR #836, opened 2023-05-12
Make Win10 Compatibility Doc "Static"

### LINKED ISSUE #830: Windows 10 Compatibility page shouldn't need to be versioned
### Page url

https://microsoft.github.io/react-native-windows/docs/win10-compat

### Problem Description

Since this page is mostly a table that is constantly updated with the ranges of Win10 min/target versions for ranges of RNW versions, this page should be treated as "static" and cleaned up.1

1. The latest version in docs should be double-checked as being accurate
2. The oldest version in versioned_docs should be replaced with the one from docs
3. All old versions inbetween should be deleted.

This will ensure that the page is the same no matter which version of the docs is being looked at.

chiaramooney:
### Why
Windows 10 Compatibility doc page should remain the same no matter which version docs the user is looking at. 

Adjust docs for cutting new version of website to ensure this page remains "static". 

Resolves #830 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/react-native-windows-samples/pull/836)


marlenecota:
might make more sense to move the page out of  versioned_docs into pages/en. This way we don't make a new copy every time and don't have to update old versions.

chiaramooney:
> might make more sense to move the page out of versioned_docs into pages/en. This way we don't make a new copy every time and don't have to update old versions.

@jonthysell do you have thoughts on this? Only concern I have is that the content would then live as one of the main tabs along side "docs" and "resources" instead of in the sidebar. Is this a layout choice we want?

website/versioned_docs/version-0.60/win10-compat.md
```diff
@@ -4,14 +4,15 @@ title: Backward compatibility with Windows versions
...
 
-_Note : There is an ongoing effort to decouple the Windows UI platform from the UWP SDK and ship out of band – [WinUI 3.0](https://github.com/microsoft/microsoft-ui-xaml/blob/master/docs/roadmap.md#winui-3-q4-2019---2020). React Native for Windows will layer on top of this lifted WinUI platform once it is available at which time several of the below unsupported APIs may start lighting up._
+_Note : There is an ongoing effort to decouple the Windows UI platform from the UWP SDK and ship out of band – [WinUI 3.0](https://github.com/microsoft/microsoft-ui-xaml/blob/master/docs/roadmap.md#winui-3-q4-2019---2020). React Native for Windows will layer on top of this lifted WinUI platform once it is available at which time several of the below unsupported APIs may start lighting up. For more information on WinUI 3 support and status, see [this page](winui3.md)._
```
jonthysell, line 10: It wasn't tracked in an issue, but we want to remove this WinUI3 block from this page since the POR for Fabric is using comp instead.

website/versioned_docs/version-0.60/win10-compat.md
```diff
@@ -22,4 +23,6 @@ The following table captures the versions of Windows 10 that a React Native for
 
 | React Native Windows | Target OS Version | Minimum OS Version |
 | :--: | :-: | :-: |
-| 0.60 | **May 2019 update**<br> Version 1903 ; Build 10.0.18362.1 | **Creators Update**<br> Version 1703 ; Build 10.0.15063.468 |
+| 0.66+ | **May 2020 update**<br> Version 2004 ; Build 10.0.19041.0 | **Fall Creators Update**<br> Version 1709 ; Build 10.0.16299.91 |
```
jonthysell, line 26: Does this need to be updated due to the changes for WinUI 2.8 bumping the minimum version in 0.72?

website/versioned_docs/version-0.60/winui3.md
jonthysell, line 1: If you remove the link in the win10 page you won't need to add this file here.

jonthysell (approved):

jonthysell:
> > might make more sense to move the page out of versioned_docs into pages/en. This way we don't make a new copy every time and don't have to update old versions.
> 
> @jonthysell Jon Thysell (JAUNTY) FTE do you have thoughts on this? Only concern I have is that the content would then live as one of the main tabs along side "docs" and "resources" instead of in the sidebar. Is this a layout choice we want?

We don't **have** to add the link to the main tabs on the top, that's controlled by the `headerLinks` in `siteConfig.js`. The only real thing is static pages are JS (maybe with embedded MD like the support page) rather than an MD docs page. You'd also have to update any links **to** the new page.

website/versioned_docs/version-0.60/win10-compat.md
chiaramooney: Yes it will be! I'm going to have yajur add that change when he cuts the docs for 72.

chiaramooney:
> > > might make more sense to move the page out of versioned_docs into pages/en. This way we don't make a new copy every time and don't have to update old versions.
> > 
> > 
> > @jonthysell Jon Thysell (JAUNTY) FTE do you have thoughts on this? Only concern I have is that the content would then live as one of the main tabs along side "docs" and "resources" instead of in the sidebar. Is this a layout choice we want?
> 
> We don't **have** to add the link to the main tabs on the top, that's controlled by the `headerLinks` in `siteConfig.js`. The only real thing is static pages are JS (maybe with embedded MD like the support page) rather than an MD docs page. You'd also have to update any links **to** the new page.

Ah I see. Maybe I'll leave as a regular page for now. Since we don't expect the compatibility to change soon, but if the process becomes annoying to maintain we can put in the extra effort to convert the page to JS.

Issue #620, opened 2022-01-12
Some docs might be better served as "evergreen" static pages


jonthysell:
We have some docs pages that don't make sense to be versioned according to RNW version. It might make sense to migrate these to static pages, so they're both easier to maintain and make sure users are given the most correct information.

Potential pages:
* https://microsoft.github.io/react-native-windows/docs/supported-community-modules
* https://microsoft.github.io/react-native-windows/docs/win10-compat

PR #882, opened 2023-10-10
Remove prompt for community module issues

chrisglein:
## Description
The current recommendation for non-desktop-supporting modules is to open an issue on the module's repo.
This change also removes the list of supported modules, as it's out of date and directory is a better source anyway.

### Why
We [removed the issue template](https://github.com/microsoft/react-native-windows/commit/8016969d5c30c2dacb2cf0d89e7eafa447457269) but not the request to log an issue on our repo.

## Screenshots
image

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/react-native-windows-samples/pull/882)


chrisglein:
Would be nice to take the time to move to a static page (https://github.com/microsoft/react-native-windows-samples/issues/620), if I can just figure out a good example to model after...

jonthysell (approved):

PR #913, opened 2024-02-06
Fixed tag and props filename

anacardix:
WinUI2xVersion tag was wrong and also the props filename


jonthysell (approved):

anacardix:
@microsoft-github-policy-service agree

PR #926, opened 2024-02-25
Update getting-started.md ; need to npm install react-native 

hfnhf:
## Description

Getting started for fresh windows installations leads to a tripping point for new users who simply start with the first npx command. (npx react-native@latest init <projectName> --version "latest")  does not install react-native 

### Why
need react-native installed otherwise the second npx command [:](npxSecondCommandTrippingPoint)
```
npx react-native-windows-init --overwrite
```
has error:
Reading project name from package.json...
Reading react-native version from node_modules...
Cannot find module 'react-native/package.json'

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/react-native-windows-samples/pull/926)

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -25,12 +25,14 @@ Remember to call `react-native init` from the place you want your project direct
 npx react-native@nightly init <projectName> --version "nightly"
 ```
 
-### Navigate into this newly created directory
+### Navigate into this newly created directory and use npm to install react-native
 
 Once your project has been initialized, React Native will have created a new sub directory where all your generated files live.
+Use NPM to install react-native.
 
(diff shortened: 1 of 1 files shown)


chrisglein (changes requested):
Installing react-native through npm after creating from template shouldn't be a required step. Can you provide more detail about what you steps you took and what you hit here?

hfnhf:
Closing pull request, probably it was my own user-environment jank.
I tried default npm cache troubleshooting, but ended up making a new windows user and doing a new volta installation of node LTS and PR is not needed on the other user at this time.

PR #969, opened 2024-09-06
Update rnm-getting-started.md to recommend using 0.75

gabrieldonadel:
## Description

Update `rnm-getting-started.md` to recommend using react-native 0.75

### Why

This documentation has been outdated for quite a while now and it's still recommending react-native 0.71. Now that 0.75 is out we should recommend that instead 

## Screenshots

N / A

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/react-native-windows-samples/pull/969)

### DIFF
--- a/docs/rnm-getting-started.md
+++ b/docs/rnm-getting-started.md
@@ -5,7 +5,7 @@ title: Get Started with macOS
 
 This guide will help you get started on setting up your very first React Native for macOS app.
 
->** Latest stable version available for React Native for macOS is 0.71**
+>** Latest stable version available for React Native for macOS is 0.75**
 
 For information around how to set up:
 - React Native for iOS and Android: See [React Native Getting Started Guide](https://reactnative.dev/docs/getting-started)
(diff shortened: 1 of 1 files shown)


docs/rnm-getting-started.md
```diff
@@ -5,7 +5,7 @@ title: Get Started with macOS
 
 This guide will help you get started on setting up your very first React Native for macOS app.
 
->** Latest stable version available for React Native for macOS is 0.71**
+>** Latest stable version available for React Native for macOS is 0.75**
```
jonthysell, line 8: This won't show up on the website. You'll also need to edit (create) the 0.75 versioned doc of this file. You'll also probably want to fix the "old" new project commands (`react-native@latest init`) since that won't work anymore.

jonthysell:
+@Saadnajmi, can you take a look at this?

Saadnajmi:
Hey @gabrieldonadel Thanks for the PR. For transparency: I often forget to update this repo as it's mostly outside of my view (partially evidenced by the fact it's in react-native-**windows**-samples). We (the macOS team) had been focused on getting `0.74` shipped and `0.75` shipped faster, which meant we missed a few of the steps we usually go back and do, like:
- Update this repo
- Update the upgrade-helper

I'm currently on vacation, but I'll be sure to take a look at both next week when I'm back.

chrisglein:
Getting Started appears to be correct now given #1003

PR #991, opened 2024-11-14
NativeModuleSample based on `cpp-lib` template

jonthysell:
## Description

This PR intends to replace the existing sample at `NativeModuleSample\cppwinrt` with a new `NativeModuleSample\cpp-lib`, based on the `cpp-lib` template exposed in RNW 0.76.0.

Features of this new sample includes:
* Built-in support for supporting both New and Old Architecture app projects
* Two example apps (one Old Arch, one New Arch) to showcase the E2E works
* Upgrade from JS to TS
* Faster native builds thanks to building against the NuGet packages

### Why
The old sample is based on the Old Architecture, UWP-based C++ library template. That template is not the target of active development, while the new template is the new standard for RNW >= 0.76.0.

## Screenshots
| Old Arch Example | New Arch Example |
|:-:|:-:|
| image |  image |

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/react-native-windows-samples/pull/991)


jonthysell:
Note: I will update our native module documentation in a separate PR.

Yajur-Grover (approved):
Quick question: why don't we add the `RnwNewArch` constants to the old arch templates with the value set to `false`? Do we want the main upgrade path to be just running the new templates?

jonthysell:
> Quick question: why don't we add the `RnwNewArch` constants to the old arch templates with the value set to `false`? Do we want the main upgrade path to be just running the new templates?

Yes, the intent for the new library templates (currently just `cpp-lib`, but hopefully with a `cs-lib`) is to support both Old and New architecture, since that decision is made by the app, not the library.

If the value isn't set (aka every project not 0.76.0+'s `cpp-app` template), it gets set to `false`.

While the library does appear to set it to true in it's `ExperimentalFeatures.props` file, that file is only loaded if you open the sln file next to it (aka you just want to build the library by itself, with no app). So if you want to build just the library by itself, but for old arch, you have to change the variable there.

But in this new world where (new) libraries come with a ready to use example app, it's actually a better experience to build the example app so you can actually test that your native module works. And therefore in this example we have two example apps, so you can test both new and old architectures.

PR #1009, opened 2025-02-24
Fix steps for creating a new app to avoid incompatible versions of `react-native` and `react-native-windows` being used

Yajur-Grover:
## Description
Fix steps for creating a new app to avoid incompatible versions of `react-native` and `react-native-windows` being used

### Why
Prevents bad experience for new users of RNW.

An issue was hit where using the latest version of `react-native` (0.78) to create an app would lead to it being incompatible with the latest version of `react-native-windows`. These steps make sure that the user is using a compatible version of `react-native` for their desired version of `react-native-windows`.

## Screenshots

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/react-native-windows-samples/pull/1009)


website/versioned_docs/version-0.77/getting-started.md
```diff
@@ -26,7 +39,7 @@ Remember to call `@react-native-community/cli init` from the place you want your
 <!-- See https://www.npmjs.com/package/react-native?activeTab=versions for the RN version tags. -->
 
 ```bat
-npx --yes @react-native-community/cli@next init <projectName> --version "latest"
+npx --yes @react-native-community/cli@next init <projectName> --version "0.77.1"
```
TatianaKapos, line 42: This should be "0.77-stable" but Meta hasn't made that tag yet, just pinged them to create it

TatianaKapos: If you are manually creating a 0.77-stable app here, I'm not sure if you need to add the warning above, the app will always be 0.77-stable (and should be since it's the 0.77 documentation) and not use latest.

website/versioned_docs/version-0.77/getting-started.md
```diff
@@ -14,8 +14,21 @@ For information around how to set up React Native, see the [React Native Getting
 
 ## Create a new React Native project
 
-Remember to call `@react-native-community/cli init` from the place you want your project directory to live.
+First, we must determine the version of `react-native-windows` you wish to use. For this example, we'll say the `latest` build since that's the latest stable release. If you run the following, you'll get the concrete version of `react-native-windows`'s `latest` tag:
 
+```bat
+npm show react-native-windows@latest version
```
jonthysell, line 20: I don't think this is necessary - we only get out of sync in the delay between the when RN releases a new latest and then we release our matching latest.

Since they don't create a 0.XX-stable until **after** they release a new latest, we can't rely on that either.

I think that now that we have to use the react-native community CLI, it's easier to just call init with a hat version : `^0.77.0`

website/versioned_docs/version-0.77/getting-started.md
jonthysell, line 42:
```suggestion
npx --yes @react-native-community/cli@next init <projectName> --version "^0.77.0"
```

TatianaKapos (approved):

PR #1017, opened 2025-03-19
Fix CodeQL issues for checked-in yarn binaries

### LINKED ISSUE #1016: Resolve CodeQL alerts for checked-in `.yarn` paths in some samples
Now that CodeQL is scanning this repo, and even though I'm trying to prevent scanning the samples (as they're non-production) we're getting issues thanks to the checked-in yarn binaries in the `.yarn` folders of several samples and the security issues they bring with them.

It's hard to tell if we can remove the folders entirely (see https://github.com/yarnpkg/yarn/issues/7741) or if we'll have to try to exclude them from CodeQL scanning.

jonthysell:
## Description

This PR removes unnecessary `.yarn` folders from the repo and prevents them from being checked-in / scanned by CodeQL in the future.

### Why
Having yarn binaries checked-in is a security risk and CodeQL complains about issues in the scripts (which we can't fix, or even see, as they're obfuscated as binary files).

Removing the `.yarn` folder for older samples is not a problem, but newer samples which use Yarn 3, such as `NativeModuleSample/cpp-lib`, need the Yarn 3 binaries checked-in.

Unfortunately there's no way to get around this. Yarn refuses to acknowledge this anti-pattern and security issue (see https://github.com/yarnpkg/yarn/issues/7741) and the "solution" to require users to install Yarn 3 via `corepack` does not work in GitHub CI (see https://github.com/actions/setup-node/issues/480). I've tried every workaround combination in these two issues and none work.

Closes #1016 

## Screenshots
N/A

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/react-native-windows-samples/pull/1017)


Saadnajmi:
Just an FYI, the Node.JS Foundation voted to remove corral from the executable from Node 24 or 25 onwards

jonthysell:
> Just an FYI, the Node.JS Foundation voted to remove corral from the executable from Node 24 or 25 onwards

Do you mean corepack? That's fine, I only tried enabling it in this PR as a workaround to removing the checked-in yarn 3 cjs files. But since I couldn't get it to work in CI (it worked fine locally) the only option left is to just tell CodeQL to ignore it.

Yajur-Grover (approved):

Issue #1014, opened 2025-03-13
Clearly tag each doc page as Old Arch / New Arch, or Needs Review


jonthysell:
### Suggestion

Every page should clearly state whether it applies to RNW New Arch, Old Arch, or both, with a link to the New vs. OId Arch page.

PR #1024, opened 2025-04-18
Add Fabric Component Props under API and Separate Old/New Components Section

anupriya13:
## Description
- Separates Old and New Arch Components Sections
- Adds Fabric Component Props under New API for Windows.
It doesn't contain Android / IOS specific props.

### Why
What is the motivation for this change? Add a few sentences describing the context and overall goals of the pull request's commits.
Improve documentation

Resolves https://github.com/microsoft/react-native-windows-samples/issues/1025

Refer https://github.com/facebook/react-native-website/blob/main/docs/image.md?plain=1

## Screenshots
**Before:**

https://github.com/user-attachments/assets/94f2d14c-efef-42ea-8453-41f29ec543af

**After:** Added a new section which was missing in before "Windows Components - New Architecture"

https://github.com/user-attachments/assets/483051c1-5050-4259-8b26-25ed13765629

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/react-native-windows-samples/pull/1024)


anupriya13:
@microsoft-github-policy-service agree company="Microsoft"

Yajur-Grover:
Couple of small things:

- In the sidebar, the titles of the Old and New Arch sections could be changed to just be 'Components - New Architecture' and 'Components - Old Architecture'. The 'Windows' part is not necessary since it's the React Native for Windows website.
- Some of the components that you have put in the New Architecture section are also present in Old Architecture, which may cause some confusion for newer users of React Native Windows. Maybe it might make sense to have a list of Core components that are present in (or planned to be present in) both Old and New Architecture, and then a section for Old Arch only components, and one for New Arch only components.

website/.unbroken_exclusions
```diff
@@ -17,6 +17,7 @@
 !versioned_docs/version-0.76/native-api/*-api-windows*.md
 !versioned_docs/version-0.77/native-api/*-api-windows*.md
 !versioned_docs/version-0.78/native-api/*-api-windows*.md
+!versioned_docs/version-0.79/native-api/*-api-windows*.md
```
TatianaKapos, line 20: We don't want to cut these documents yet, RNW 0.79 isn't released yet. Instead, this should just be changed in the next documentation (main) and we will create the 0.79 version-docs as part of the release process

website/.unbroken_exclusions
```diff
@@ -31,6 +32,13 @@ File not found native-api/ReactNativeHost-api-windows.md while parsing versioned
 File not found IViewManagerCreateWithProperties while parsing versioned_docs/version-0.76/view-managers.md
 
 #fix-unbroken.js auto-generated do not edit this line or below
+File not found getting-started.md while parsing versioned_docs/version-0.79/getting-started.md
```
TatianaKapos, line 35: Same comment as above.

website/versioned_docs/version-0.79/getting-started.md
```diff
@@ -0,0 +1,135 @@
+---
+id: version-0.79-getting-started
```
TatianaKapos, line 2: Same comment as above. this should just change the getting-started document in the docs folder

TatianaKapos (changes requested):
We aren't ready to cut the 0.79 documentation, any changes/added documentation should be in the docs folder

anupriya13:
> Couple of small things:
> 
> * In the sidebar, the titles of the Old and New Arch sections could be changed to just be 'Components - New Architecture' and 'Components - Old Architecture'. The 'Windows' part is not necessary since it's the React Native for Windows website.
> * Some of the components that you have put in the New Architecture section are also present in Old Architecture, which may cause some confusion for newer users of React Native Windows. Maybe it might make sense to have a list of Core components that are present in (or planned to be present in) both Old and New Architecture, and then a section for Old Arch only components, and one for New Arch only components.

1. Sure, will update the titles
2. The reason I kept it as different is because each page of old and new would have different props, and when we want to depreciate the old one it would be easier. Helps us track exactly what props are implemented. Also, the old arch documentation is not complete, wouldn't want to add effort in updating that.
Hence, keeping these 2 separate only for now. We can update in future after discussing with PMs.

website/.unbroken_exclusions
anupriya13: If we don't add this then it's showing some PR build failures.. Is it fine? I have removed rest of the stuff as well. Please review.

Yajur-Grover:
> 2. The reason I kept it as different is because each page of old and new would have different props, and when we want to depreciate the old one it would be easier. Helps us track exactly what props are implemented. Also, the old arch documentation is not complete, wouldn't want to add effort in updating that.
>    Hence, keeping these 2 separate only for now. We can update in future after discussing with PMs.

Okay, maybe we can apply the suggestion from https://github.com/microsoft/react-native-windows-samples/issues/1014 then, and mark each page depending on whether it is applicable to Old Arch, New Arch or both.

chiaramooney (changes requested):
Blocking until design decision made about if we want to document all components instead of Windows-only.

PR #1051, opened 2025-07-03
Update documentation to reflect New Architecture default

### LINKED ISSUE #1046: Create Documentation for "Can I Migrate My Paper App To Fabric?"
### Suggestion

As we move toward Fabric as the default, more developers will be wondering if Fabric is now ready for them to migrate their Paper RNW app to Fabric. While Fabric has reached over 80% parity, it has not reached full parity and there are many popular community modules that we don't yet support. We should create detailed documentation (on the website) explaining to developers what is/isn't supported on Fabric (this can include links to GitHub queries) and a list of modules supported/no supported on Fabric (can be a reactnative.directory query). 

The page should also encourage developers to file issues for props/modules they need support for that we aren't currently tracking. 

We've contemplated proposed a script which could search through an app's dependencies/source to determine if an app can move to Fabric yet. It's unclear if this is technically possible. Idea would require some technical investigation.

We want our documentation to be as clear as possible. We want to avoid developers trying to upgrade to Fabric, hitting unexpected errors/blocks, and then repeating this process on loop for each React Native Windows version that's released.

### Location

Advanced Topics (Windows), with callout to the page in Getting Started/Upgrading

jonthysell:
## Description

This PR updates the documentation to reflect the New Architecture default for new applications in 0.80, and also refreshes the documentation on the differences between the architectures for RNW.

### Why
We are switching the default for new projects in 0.80. We are also directing more folks to the "New vs. Old Architecture" page both within the docs and from outside.

Resolves #1046

## Screenshots
N/A

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/react-native-windows-samples/pull/1051)

### DIFF
--- a/docs/new-architecture.md
+++ b/docs/new-architecture.md
@@ -5,101 +5,143 @@ title: New vs. Old Architecture
 
 ![Architecture](https://img.shields.io/badge/architecture-new_&_old-green)
 
-React Native's [New Architecture](https://reactnative.dev/architecture/landing-page) has become the default with [version 0.76](https://reactnative.dev/blog/2024/10/23/the-new-architecture-is-here), bringing many framework improvements including the advanced rendering system [Fabric](https://reactnative.dev/architecture/fabric-renderer). While React Native for Windows isn't quite yet ready to make the New Architecture the default, we've been hard at work and are excited to offer a sneak peek into adopting it on Windows.
+React Native's [New Architecture](https://reactnative.dev/architecture/landing-page) brings many framework improvements including the advanced [Fabric](https://reactnative.dev/architecture/fabric-renderer) rendering system. It greatly improves the speed and responsiveness of React Native apps over the previous *Legacy* or *Old Architecture* and its *Paper* renderer. The New Architecture was enabled by default for the *iOS* and *Android* platforms in [react-native@0.76.0](https://reactnative.dev/blog/2024/10/23/the-new-architecture-is-here).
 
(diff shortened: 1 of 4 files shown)


docs/new-architecture.md
```diff
@@ -5,101 +5,143 @@ title: New vs. Old Architecture
...
+React Native for Windows also supports the New Architecture for the *Windows* platform. It was first available as an opt-in preview in [react-native-windows@0.76.0](https://devblogs.microsoft.com/react-native/2025-01-29-new-architecture-on-0-76-0-77/) and became the default for *new* app projects in `react-native-windows@0.80.0`. For more information on the differences between the architectures on Windows, see [New vs. Old Architecture Differences](#new-vs-old-architecture-differences).
 
-## From UWP to WinAppSDK
+> **Important:** Nearly all future investments to React Native (and React Native for Windows) will be toward the New Architecture, and we highly recommend that all projects migrate as soon as possible. React Native (and therefore React Native for Windows) will eventually deprecate, then delete, support for the Old Architecture.
```
chiaramooney, line 12: Meta just announced they are freezing their Paper implementation (https://github.com/reactwg/react-native-new-architecture/discussions/290). Up to you if you want to mention that here.

chiaramooney (approved):

docs/new-architecture.md
jonthysell: I'l' make the changes in a follow-up.

